PR #94, opened 2021-09-07
feat: real time syncing into a sync strategy class.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Spilt the real time syncing from the `SyncHandler` into its own class after adding the necessary interfaces and classes to support other sync strategies.

This change will enable shadow manager to support multiple sync strategies.

Added and updated unit tests and integration tests.

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/NucleusLaunchUtils.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/NucleusLaunchUtils.java
@@ -17,6 +17,7 @@ import com.aws.greengrass.shadowmanager.ShadowManagerDatabase;
 import com.aws.greengrass.shadowmanager.exception.RetryableException;
 import com.aws.greengrass.shadowmanager.sync.IotDataPlaneClientFactory;
 import com.aws.greengrass.shadowmanager.sync.SyncHandler;
+import com.aws.greengrass.shadowmanager.sync.strategy.RealTimeSyncStrategy;
 import com.aws.greengrass.testcommons.testutilities.GGServiceTestUtil;
 import com.aws.greengrass.util.RetryUtils;
 import org.junit.jupiter.api.io.TempDir;
@@ -29,6 +30,7 @@ import java.util.Collections;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.lenient;
@@ -52,18 +54,72 @@ public class NucleusLaunchUtils extends GGServiceTestUtil {
     @Mock
     ShadowManagerDatabase mockShadowManagerDatabase;
 
+    @Deprecated()
     public void startNucleusWithConfig(String configFile) throws InterruptedException {
         startNucleusWithConfig(configFile, State.RUNNING, false, false, false);
     }
 
+    @Deprecated()
     void startNucleusWithConfig(String configFile, boolean mockCloud, boolean mockDao) throws InterruptedException {
         startNucleusWithConfig(configFile, State.RUNNING, false, mockCloud, mockDao);
     }
 
+    @Deprecated()
     void startNucleusWithConfig(String configFile, State expectedState, boolean mockDatabase) throws InterruptedException {
         startNucleusWithConfig(configFile, expectedState, mockDatabase, false, true);
     }
 
+    void startNucleusWithConfig(NucleusLaunchUtilsConfig config) throws InterruptedException {
+        CountDownLatch shadowManagerRunning = new CountDownLatch(1);
+        AtomicBoolean isSyncMocked = new AtomicBoolean(false);
+        kernel.parseArgs("-r", rootDir.toAbsolutePath().toString(), "-i",
+                getClass().getResource(config.getConfigFile()).toString());
+        listener = (GreengrassService service, State was, State newState) -> {
+            if (service.getName().equals(ShadowManager.SERVICE_NAME) && service.getState().equals(config.getExpectedState())) {
+                shadowManagerRunning.countDown();
+                shadowManager = (ShadowManager) service;
+            }
+        };
+        kernel.getContext().addGlobalStateChangeListener(listener);
+        kernel.getContext().put(MqttClient.class, mqttClient);
+        if (config.isMqttConnected()) {
+            // assume we are always connected
+            lenient().when(mqttClient.connected()).thenAnswer(invocation -> isSyncMocked.get());
+        }
+
+
+        if (config.isMockDatabase()) {
+            kernel.getContext().put(ShadowManagerDatabase.class, mockShadowManagerDatabase);
+            kernel.getContext().put(AuthorizationHandlerWrapper.class, mockAuthorizationHandlerWrapper);
+        }
+        if (config.isMockCloud()) {
+            kernel.getContext().put(IotDataPlaneClientFactory.class, iotDataPlaneClientFactory);
+        }
+        if (config.isMockDao()) {
+            kernel.getContext().put(ShadowManagerDAOImpl.class, dao);
+        }
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        RealTimeSyncStrategy realTimeSyncStrategy = (RealTimeSyncStrategy) syncHandler.getOverallSyncStrategy();
+        ExecutorService es = kernel.getContext().get(ExecutorService.class);
+        // set retry config to only try once so we can test failures earlier
+        kernel.launch();
+
+        assertTrue(shadowManagerRunning.await(TEST_TIME_OUT_SEC, TimeUnit.SECONDS));
+
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                .maxAttempt(1)
+                .maxRetryInterval(Duration.ofSeconds(1))
+                .retryableExceptions(Collections.singletonList(RetryableException.class))
+                .build();
+        RealTimeSyncStrategy syncStrategy = new RealTimeSyncStrategy(es, realTimeSyncStrategy.getRetryer(), retryConfig);
+        syncHandler.setOverallSyncStrategy(syncStrategy);
+        isSyncMocked.set(true);
+        shadowManager.startSyncingShadows();
+
+        kernel.getContext().put(RealTimeSyncStrategy.class, syncStrategy);
+    }
+
+    @Deprecated()
     void startNucleusWithConfig(String configFile, State expectedState, boolean mockDatabase, boolean mockCloud,
                                 boolean mockDao) throws InterruptedException {
         CountDownLatch shadowManagerRunning = new CountDownLatch(1);
@@ -90,16 +146,7 @@ public class NucleusLaunchUtils extends GGServiceTestUtil {
         if (mockDao) {
             kernel.getContext().put(ShadowManagerDAOImpl.class, dao);
         }
-        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
-        ExecutorService es = kernel.getContext().get(ExecutorService.class);
-
         // set retry config to only try once so we can test failures earlier
-        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
-                .maxAttempt(1)
-                .maxRetryInterval(Duration.ofSeconds(1))
-                .retryableExceptions(Collections.singletonList(RetryableException.class))
-                .build();
-        kernel.getContext().put(SyncHandler.class, new SyncHandler(syncHandler.getSyncQueue(), es, retryConfig));
         kernel.launch();
 
         assertTrue(shadowManagerRunning.await(TEST_TIME_OUT_SEC, TimeUnit.SECONDS));
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/NucleusLaunchUtils.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/NucleusLaunchUtils.java
@@ -76,8 +76,8 @@ public class NucleusLaunchUtils extends GGServiceTestUtil {
                 getClass().getResource(config.getConfigFile()).toString());
         listener = (GreengrassService service, State was, State newState) -> {
             if (service.getName().equals(ShadowManager.SERVICE_NAME) && service.getState().equals(config.getExpectedState())) {
-                shadowManagerRunning.countDown();
                 shadowManager = (ShadowManager) service;
+                shadowManagerRunning.countDown();
             }
         };
         kernel.getContext().addGlobalStateChangeListener(listener);
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -91,6 +91,7 @@ public class ShadowManager extends PluginService {
     private final UpdateThingShadowRequestHandler updateThingShadowRequestHandler;
     private final GetThingShadowRequestHandler getThingShadowRequestHandler;
     private final IotDataPlaneClientWrapper iotDataPlaneClientWrapper;
+    @Getter(AccessLevel.PACKAGE)
     private final SyncHandler syncHandler;
     private final CloudDataClient cloudDataClient;
     private final MqttClient mqttClient;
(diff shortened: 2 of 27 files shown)


randolphtan121:
This change allows customers to supply their own sync strategies correct? I see a unit test mocking sync strategies but we should definitely add an integration test to capture this feature with a dummy supplied sync strategy.

Nit Comment: Update the description of pull request to say that customers can supply their own sync strategies as well.

randolphtan121 (commented):
This change allows customers to supply their own sync strategies correct? I see a unit test mocking sync strategies but we should definitely add an integration test to capture this feature with a dummy supplied sync strategy.

Nit Comment: Update the description of pull request to say that customers can supply their own sync strategies as well.

src/integrationtests/java/com/aws/greengrass/integrationtests/NucleusLaunchUtils.java
```diff
@@ -52,18 +54,72 @@
     @Mock
     ShadowManagerDatabase mockShadowManagerDatabase;
 
+    @Deprecated()
```
randolphtan121, line 57: Are we not cleaning these up because they are being used across all the unit tests?

nikkhilmuthye: Yes. I will clean them up in a later PR since these are used in all the integration tests. Didn't want to make the PR any bigger.

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -243,7 +243,15 @@ private void handleUpdate(MqttMessage message) {
...
                 .log("Received cloud update request");
-        syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload());
+        try {
+            syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.atError()
+                    .setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code())
+                    .setCause(e)
+                    .log("Failed to add local update");
```
youtuyy, line 253: nit: the error message doesn't seem very clear to me.

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -258,7 +266,15 @@ private void handleDelete(MqttMessage message) {
...
                 .log("Received cloud delete request");
-        syncHandler.pushLocalDeleteSyncRequest(thingName, shadowName, message.getPayload());
+        try {
+            syncHandler.pushLocalDeleteSyncRequest(thingName, shadowName, message.getPayload());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.atError()
+                    .setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code())
+                    .setCause(e)
+                    .log("Failed to add local delete");
```
youtuyy, line 276: same

youtuyy (commented):
Can you also update README with the new strategy config?

src/main/java/com/aws/greengrass/shadowmanager/model/ShadowDocument.java
```diff
@@ -65,8 +78,10 @@ public ShadowDocument(byte[] documentBytes) throws IOException, InvalidRequestPa
                 .orElseThrow(() ->
                         new InvalidRequestParametersException(ErrorMessage
                                 .createInvalidPayloadJsonMessage("")));
-        // Validate the payload schema
-        JsonUtil.validatePayloadSchema(documentJson);
+        if (validate) {
```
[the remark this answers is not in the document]
nikkhilmuthye: There are instances where we only need to get the version number from a JSON byte array. We get that as a response from the cloud when we update it. So we don't need to validate the entire JSON there.

nikkhilmuthye:
> Can you also update README with the new strategy config?

I added that in the next PR -> https://github.com/aws-greengrass/aws-greengrass-shadow-manager/pull/95

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+        try {
+            syncQueue.put(request);
+
+            // the above put call will block. If syncing is stopped while waiting but after the put call succeeds then
+            // remove the request we just added
+            if (!syncing.get()) {
+                syncQueue.remove(request);
+            }
+        } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
nikkhilmuthye: We want to propagate that up to the caller so that we can log it there and send the appropriate messages over PubSub if needed.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -291,6 +294,18 @@ protected void install() {
...
                 });
+
+        Topics strategyTopics = config.lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_STRATEGY_TOPIC);
+        strategyTopics.subscribe((why, newv) -> {
+                    Strategy strategy;
+                    if (WhatHappened.removed.equals(why) || newv == null) {
+                        strategy = DEFAULT_STRATEGY;
+                    } else {
+                        // Get the correct sync strategy configuration from the POJO.
+                        strategy = Strategy.fromPojo(strategyTopics.toPOJO());
```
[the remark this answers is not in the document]
nikkhilmuthye: I think I designed it to default to real time syncing if there is some error in the config. But I agree. I will handle the illegal configuration and error out the service in that situation.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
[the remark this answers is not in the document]
youtuyy: Since the caller will handle this, should we not catch it here? or should we re throw it maybe as a custom exception?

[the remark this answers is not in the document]
nikkhilmuthye: We are reinterrupting the the thread. That should be enough. We don't need to throw any other exception here.

youtuyy (approved):

randolphtan121 (approved):

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -242,8 +242,16 @@ private void handleUpdate(MqttMessage message) {
...
-                .log("Received cloud update request");
-        syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload());
+                .log("Received cloud update sync request");
+        try {
+            syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.atError()
+                    .setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code())
+                    .setCause(e)
```
rbattle, line 252: I don't think it makes sense to set the cause here - printing an InterruptedException is not generally useful. You could include in the error log that it was interrupted while adding the local sync request

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -0,0 +1,111 @@
...
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Future;
+
+import static com.aws.greengrass.shadowmanager.model.LogEvents.SYNC;
+
+public class BaseSyncStrategy {
```
rbattle, line 27: abstract?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.shadowmanager.sync.strategy;
+
+import com.aws.greengrass.shadowmanager.sync.Retryer;
+import com.aws.greengrass.shadowmanager.sync.model.SyncContext;
+import com.aws.greengrass.shadowmanager.sync.model.SyncRequest;
+
+public class PeriodicSyncStrategy extends BaseSyncStrategy implements SyncStrategy {
```
rbattle, line 12: needs a class level comment describing what this does

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
```diff
@@ -0,0 +1,248 @@
...
+import software.amazon.awssdk.aws.greengrass.model.ConflictError;
+import software.amazon.awssdk.services.iotdataplane.model.ConflictException;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_SHADOW_NAME_KEY;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_THING_NAME_KEY;
+
+public class RealTimeSyncStrategy extends BaseSyncStrategy implements SyncStrategy {
```
rbattle, line 26: class level comment

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * Starts syncing the shadows based on the strategy.
+     *
+     * @param context         an context object for syncing
+     * @param syncParallelism number of threads to use for syncing
+     */
+    @Override
+    public void startSync(SyncContext context, int syncParallelism) {
+        synchronized (lifecycleLock) {
+            if (syncing.compareAndSet(false, true)) {
```
rbattle, line 70: nit: could this be moved to the Base class (especially if it is made abstract) and then you have a `doStop` and `doStart` that don't need to do this synchronization and compare and set?

e.g.: 
```
public void startSync(...) {
        synchronized (lifecycleLock) {
            if (syncing.compareAndSet(false, true)) {
              doStart(context, syncParallelism);
            } else {
               logger.atDebug(SYNC_EVENT_TYPE).log("Already started syncing");
            }
    }
}

abstract void doStart();
```

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
```diff
@@ -0,0 +1,248 @@
...
+     * We cannot support an unbounded queue as that will lead to memory pressure - instead requests need to be
+     * throttled.
+     * <p/>
+     * Synchronized so that there is at most only one put in progress waiting to be added if queue is full
+     *
+     * @param request request the request to add.
+     * @throws InterruptedException if the thread is interrupted while enqueuing data
+     */
+    @Override
+    public void putSyncRequest(SyncRequest request) throws InterruptedException {
```
rbattle, line 127: does this method actually throw InterruptedException anymore? Since you caught it and set the interrupt on the thread I don't think it does.

Does anything calling this check if `Thread.currentThread().isInterrupted()`?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
```diff
@@ -0,0 +1,248 @@
...
+                            .cause(e)
+                            .addKeyValue(LOG_THING_NAME_KEY, request.getThingName())
+                            .addKeyValue(LOG_SHADOW_NAME_KEY, request.getShadowName())
+                            .log("Skipping sync request");
+                    request = syncQueue.take();
+                }
+            } while (!Thread.currentThread().isInterrupted());
+        } catch (InterruptedException e) {
+            logger.atWarn(SYNC_EVENT_TYPE).log("Interrupted while waiting for sync requests");
+            Thread.currentThread().interrupt();
```
rbattle, line 244: not sure that this is strictly necessary since the thread is going to exit anyway after catching the exception

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/SyncStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    AtomicBoolean syncing = new AtomicBoolean(false);
+
+    /**
+     * Starts syncing the shadows based on the strategy.
+     *
+     * @param context         an context object for syncing
+     * @param syncParallelism number of threads to use for syncing
+     */
+    void startSync(SyncContext context, int syncParallelism);
```
rbattle, line 28: minor nit: you could just call this method `start` and the other `stop` - sync is redundant here

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/model/Strategy.java
```diff
@@ -0,0 +1,63 @@
...
+            switch (entry.getKey()) {
+                case "type":
+                    strategy.type(StrategyType.fromCode(Coerce.toString(entry.getValue())));
+                    break;
+                case "delay":
+                    strategy.delay(Coerce.toLong(entry.getValue()));
+                    break;
+                default:
+                    throw new InvalidConfigurationException(String.format("Unexpected key in configuration: %s",
+                            entry.getKey()));
```
rbattle, line 58: does it matter if we have an unexpected key, or does it only matter if we don't get a valid type and/or delay?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/model/Strategy.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public static Strategy fromPojo(Map<String, Object> pojo) throws InvalidConfigurationException {
+        StrategyBuilder strategy = Strategy.builder();
+        for (Map.Entry<String, Object> entry : pojo.entrySet()) {
+            switch (entry.getKey()) {
+                case "type":
+                    strategy.type(StrategyType.fromCode(Coerce.toString(entry.getValue())));
+                    break;
+                case "delay":
+                    strategy.delay(Coerce.toLong(entry.getValue()));
```
rbattle, line 54: if it's not a long, this will return 0 - is that ok?

rbattle (commented):
I've still got to go through the tests - this looks like a good refactor.

I have some questions about the InterruptedException in a few places

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
nikkhilmuthye: Yeah. I agree. I refactored the code. Nothing is really throwing `InterruptedException` here.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/model/Strategy.java
nikkhilmuthye: I don't think we want to add any min/max validation here right? Apart from that, I feel it should be on the customer to set the correct delay.

rbattle: what does a delay of 0 mean? or a delay of -1?

nikkhilmuthye: For real time syncing, the delay field is not used. For periodic, it would just make it into a less efficient real time sync since it will schedule a new thread every second.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/RealTimeSyncStrategy.java
nikkhilmuthye: Right. I didn't change this function. I just refactored it from `SyncHandler`. Removed it now.

rbattle (approved):

youtuyy (approved):

randolphtan121 (approved):

PR #95, opened 2021-09-15
feat: add periodic sync strategy.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Add a new sync strategy to sync shadows periodically. Customers can specify the interval at which `ShadowManager` should sync shadows.
Whenever a local shadow is updated, that sync requests will get queued up until it is time for the sync strategy to sync up all these requests. It is same for the cloud updates. The sync strategy will queue them up as well.
When the sync loop will run, the sync strategy will run until all the sync requests have been executed.

Customers want the ability to not sync the shadows in real time based on different use cases. This helps them maintain their bandwidth consumptions.

Added new unit and integrations tests.

**Checklist:**
 

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -147,8 +150,9 @@ void stopSync() {
                 doStop();
 
                 int remaining = syncQueue.size();
-                syncQueue.clear();
 
+                // Not clearing the queue since we need it if the customer updates the sync strategy on the fly. The
+                //queue will be transferred to the new sync strategy.
```
rbattle, line 155: nit: formatting

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -82,6 +137,79 @@ public void clearSyncQueue() {
...
-        return 0;
+        return syncQueue.remainingCapacity();
+    }
+
+    /**
+     * Take and execute items from the sync queue. This is intended to be run in a separate thread.
+     * This will work on all the sync requests in the queue until it's empty.
+     */
+    @SuppressWarnings({"PMD.CompareObjectsWithEquals", "PMD.AvoidCatchingGenericException"})
+    private void syncLoop() {
```
rbattle, line 148: is this method any different from the RealTime? 
Seems like the only difference is that it calls `poll` instead of `take`

Could it go in the base class where you give it a `Supplier<SyncRequest>`?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -28,25 +77,30 @@ public PeriodicSyncStrategy(Retryer retryer) {
...
-
+        logger.atInfo(SYNC_EVENT_TYPE).kv("interval", interval).log("Start periodic syncing");
+        this.scheduledFuture = syncExecutorService
+                .scheduleAtFixedRate(this::syncLoop, 0, interval, TimeUnit.SECONDS);
     }
 
     @Override
     void doStop() {
-
+        logger.atDebug(SYNC_EVENT_TYPE).log("Cancel {} periodic sync thread(s)", syncThreads.size());
```
rbattle, line 92: nit: log the start and stop at the same log level. log is at Info, stop is at Debug. Maybe add a separate info log that you that says "Stop periodic syncing")?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -28,25 +77,30 @@ public PeriodicSyncStrategy(Retryer retryer) {
...
+        }
     }
 
     /**
      * Stops the syncing of shadows.
      */
     @Override
     public void stop() {
-
+        super.stopSync();
```
rbattle, line 103: If you made the `BaseSyncStrategy` implement `SyncStrategy` then could `start(context, syncParallelism)`, `stop()` and `putSyncRequest(SyncRequest request)` go in the base class?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -82,6 +137,79 @@ public void clearSyncQueue() {
...
+    /**
+     * Take and execute items from the sync queue. This is intended to be run in a separate thread.
+     * This will work on all the sync requests in the queue until it's empty.
+     */
+    @SuppressWarnings({"PMD.CompareObjectsWithEquals", "PMD.AvoidCatchingGenericException"})
+    private void syncLoop() {
+        logger.atInfo(SYNC_EVENT_TYPE).log("Start processing sync requests");
+        RetryUtils.RetryConfig retryConfig = this.retryConfig;
+        SyncRequest request = syncQueue.poll();
+        while (request != null) {
```
rbattle, line 152: You schedule a thread every X seconds that runs this loop. If this loop takes >= X seconds to process,  you'll have multiple threads running the syncLoop. 

Should you have a check that this isn't already running?

README.md
```diff
@@ -13,6 +13,9 @@ Manifests:
       aws.greengrass.ShadowManager
   - aws.greengrass.ShadowManager:
       Configuration:
+        strategy:
+          type: periodic // default realTime
+          delay: 300 // in seconds
```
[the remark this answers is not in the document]
nikkhilmuthye: Not right now. I don't believe it is supported in any existing components like FSS or telemetry. I can change the name to `delaySec` to make it clearer.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
nikkhilmuthye: Apart from that difference, In the real time sync, the while loop exits when the thread is interrupted. In periodic sync, it exists when the sync queue is empty. So I think we can keep it as is for now.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
nikkhilmuthye: They are ever so slightly different. I have put the common functionality in the base class as much as possible.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
nikkhilmuthye: Yeah. I agree. I thought that function was surrounded by a lock. But it isn't. Adding a check.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
rbattle: why does the periodic sync not exit when the thread is interrupted?

Does that get it into an infinite loop? The next time it the syncQueue is polled, won't that throw an InterruptedException because the current thread is interrupted? The reentrant lock in the syncQueue calls `lockInterruptibly` which will throw

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -164,4 +168,34 @@ void stopSync() {
...
+                    .addKeyValue(LOG_THING_NAME_KEY, request.getThingName())
+                    .addKeyValue(LOG_SHADOW_NAME_KEY, request.getShadowName())
+                    .log("Syncing is stopped. Ignoring sync request");
+            return;
+        }
+        try {
+            logger.atInfo(SYNC_EVENT_TYPE)
+                    .addKeyValue(LOG_THING_NAME_KEY, request.getThingName())
+                    .addKeyValue(LOG_SHADOW_NAME_KEY, request.getShadowName())
+                    .log("Adding new sync request");
```
rbattle, line 184: nit: should probably be debug here - if they get a ton of sync items this will spam logs

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
rbattle: what's different?
putSyncRequest looks to be same implementation in both Periodic and Realtime
start is the same
stop is the same

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
nikkhilmuthye: Good point. I will update the code to exit if the thread was interrupted.

src/main/java/com/aws/greengrass/shadowmanager/sync/RequestBlockingQueue.java
```diff
@@ -369,9 +369,10 @@ public SyncRequest poll(long timeout, TimeUnit unit) throws InterruptedException
      * Take an item from the queue. Returns immediately if no item is available.
      *
      * @return the request at the head of the queue or null if no items are available.
+     * @throws InterruptedException if the thread is interrupted while waiting
      */
-    public SyncRequest poll() {
-        lock.lock();
+    public SyncRequest poll() throws InterruptedException {
```
rbattle, line 374: why make this change? It was intentionally copying the java BlockingQueue interface  (but not implementing since it's not a collection) where `poll` is not interruptible unless you specify a timeout.

It returns immediately if there's no data, so there's no need to wait or interrupt.

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
rbattle: There's no difference here:

`start` in `PeriodicSyncStrategy` (line 81) is:
```
    public void start(SyncContext context, int syncParallelism) {
        super.startSync(context, syncParallelism);
    }

```

`start` in `RealTimeSyncStrategy` (line 70) is :
```
@Override
    public void start(SyncContext context, int syncParallelism) {
        super.startSync(context, syncParallelism);
    }
 ```

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
rbattle: I think I gave you bad advice - the original `poll` should not have thrown an exception since it was using the regular lock (not lockInterruptibly). I must have looked at the other implemented method when checking before.

`poll` is intended to not block so you don't need to worry about it interrupting, whereas `poll(timeout)` does block.

If poll doesn't throw an interrupted exception, then the request will either be null or the thread will be interrupted and exit

rbattle (approved):

PR #97, opened 2021-10-08
fix: increment shadow version on shadow delete.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Update the shadow version after a local shadow has been deleted. Also update the last synced local and cloud shadow version after either of them have been deleted and synced with each other.

IoT Device service increments the version of the shadow on a delete. GG needs to make sure that happens in the local shadow service as well.

Added unit tests and updates existing unit and integration tests.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/sync/model/FullShadowSyncRequest.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/sync/model/FullShadowSyncRequest.java
@@ -118,7 +118,7 @@ public class FullShadowSyncRequest extends BaseSyncRequest {
             if (isFirstSync(syncInformation.get())) {
                 handleFirstCloudSync(localShadowDocument.get());
             } else {
-                handleLocalDelete(localShadowDocument.get(), syncInformation.get());
+                handleLocalDelete(syncInformation.get());
             }
             return;
         }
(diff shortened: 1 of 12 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/model/FullShadowSyncRequest.java
```diff
@@ -259,15 +265,21 @@ private void handleFirstLocalSync(@NonNull ShadowDocument cloudShadowDocument, l
...
+        // Since the local shadow has been deleted, we need get the deleted shadow version from the DAO.
+        long localShadowVersion = context.getDao().getDeletedShadowVersion(getThingName(), getShadowName())
+                .orElse(syncInformation.getLocalVersion() + 1);
         context.getDao().updateSyncInformation(SyncInformation.builder()
-                .localVersion(localShadowDocument.getVersion())
-                .cloudVersion(syncInformation.getCloudVersion())
+                .localVersion(localShadowVersion)
+                // If the cloud shadow was deleted, then the last synced version might be 1 higher than the last version
+                // that was synced.
+                // If the device was offline for a long time and the cloud shadow was deleted multiple times in that
```
youtuyy, line 279: what does it mean by "was deleted multiple times"? what would the cloud version be in this case? does it keep incrementing?  
btw is there iot doc on how cloud deletion results in version increment?

nikkhilmuthye: So if the cloud shadow was created and deleted multiple times, the version will keep incrementing for each update and each delete. So if the cloud shadow version was 10 and the device goes offline and the cloud shadow was updated 4 times and deleted 2 times, the version of the cloud shadow would now be 16.
Here is the [link](https://docs.aws.amazon.com/iot/latest/developerguide/device-shadow-data-flow.html#deleting-thing-data) to the documentation. There is a link in this section which talks about the version increment.

youtuyy (approved):

PR #103, opened 2021-11-22
feat: add sync directionality

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Adds the ability to configure Shadow Manager to sync the shadows in a particular direction. With a configuration change, Shadow Manager will either bidirectionality sync shadows or in either one of the directions (i.e. either only from the device to the cloud or from cloud to device).

If the Shadow Manager is configured to be synced from the device to the cloud only, on first launch or after a reconnection, the cloud shadow will be overwritten to have the same content as what the local shadow has.
If the Shadow Manager is configured to be synced from the cloud to the device only, on first launch or after a reconnection, the local shadow will be overwritten to have the same content as what the cloud shadow has.

Added a new configuration field to enable customers to specify in which direction the shadows should be synced. There are numerous use cases where the device shadows only need to be synced. This will avoid unnecessary connections to the cloud if the syncing from cloud to device is not needed. 

Added unit tests and integration tests.

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncDirectionalityTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncDirectionalityTest.java
@@ -7,6 +7,7 @@ package com.aws.greengrass.integrationtests;
 
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.impl.config.LogConfig;
+import com.aws.greengrass.shadowmanager.ShadowManager;
 import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
 import com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler;
 import com.aws.greengrass.shadowmanager.model.ShadowDocument;
@@ -113,8 +114,7 @@ class SyncDirectionalityTest extends NucleusLaunchUtils {
         request.setPayload(localShadowContentV1.getBytes(UTF_8));
 
         updateHandler.handleRequest(request, "DoAll");
-        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
-        assertThat(() -> ((RealTimeSyncStrategy) syncHandler.getOverallSyncStrategy()).getSyncQueue().size(), eventuallyEval(is(0)));
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
 
         assertThat("sync info exists", () -> syncInfo.get().isPresent(), eventuallyEval(is(true)));
 
@@ -147,8 +147,12 @@ class SyncDirectionalityTest extends NucleusLaunchUtils {
                 .mockCloud(true)
                 .resetRetryConfig(false)
                 .build());
-        assertThat(() -> kernel.getContext().get(RealTimeSyncStrategy.class).getSyncQueue().size(), eventuallyEval(is(0)));
-        TimeUnit.SECONDS.sleep(5);
+        shadowManager.startSyncingShadows(ShadowManager.StartSyncInfo.builder().build());
+        // There is a race condition which can cause us to check queue is empty before having inserted any full sync
+        // requests in it. This check avoids that.
+        assertNotEmptySyncQueue(RealTimeSyncStrategy.class);
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
+
         UpdateThingShadowRequestHandler updateHandler = shadowManager.getUpdateThingShadowRequestHandler();
 
         // update local shadow
@@ -158,9 +162,9 @@ class SyncDirectionalityTest extends NucleusLaunchUtils {
         request.setPayload(localShadowContentV1.getBytes(UTF_8));
 
         updateHandler.handleRequest(request, "DoAll");
-        assertThat(() -> kernel.getContext().get(RealTimeSyncStrategy.class).getSyncQueue().size(), eventuallyEval(is(0)));
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
 
-        assertThat("local shadow exists", localShadow.get().isPresent(), is(true));
+        assertThat("local shadow exists", () -> localShadow.get().isPresent(), eventuallyEval(is(true)));
         ShadowDocument shadowDocument = localShadow.get().get();
         // remove metadata node and version (JsonNode version will fail a comparison of long vs int)
         shadowDocument = new ShadowDocument(shadowDocument.getState(), null, null);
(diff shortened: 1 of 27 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/SyncDirectionalityTest.java
```diff
@@ -0,0 +1,238 @@
...
+        when(iotDataPlaneClientFactory.getIotDataPlaneClient().getThingShadow(any(GetThingShadowRequest.class)))
+                .thenThrow(ResourceNotFoundException.class);
+
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile("sync_directionality_fromcloudonly.yaml")
+                .mockCloud(true)
+                .resetRetryConfig(false)
+                .build());
+        assertThat(() -> kernel.getContext().get(RealTimeSyncStrategy.class).getSyncQueue().size(), eventuallyEval(is(0)));
+        TimeUnit.SECONDS.sleep(5);
```
rbattle, line 151: is there a more deterministic way of waiting in these tests?

src/main/java/com/aws/greengrass/shadowmanager/sync/model/Direction.java
```diff
@@ -0,0 +1,15 @@
...
+
+package com.aws.greengrass.shadowmanager.sync.model;
+
+/**
+ * Enum to indicate the shadow sync directionality.
+ */
+public enum Direction {
+    BIDIRECTIONAL,
+    FROMDEVICEONLY,
+    FROMCLOUDONLY
```
rbattle, line 14: Few things here
1. I don't think you need the "ONLY" suffix
2. Should they have a "_" to separate words?
3. What should the directional reference be - source or the target:
FROM_DEVICE or TO_CLOUD?
FROM_CLOUD or TO_DEVICE?

When I think of the direction, I've been thinking of it in terms of where the data ends up - which is the opposite of what you have. 

Worth talking to @FThompsonAWS ?

rbattle (commented):
looks pretty good - lots of small comments.

It is nice the syncing abstraction doesn't need much change to support this feature

src/main/java/com/aws/greengrass/shadowmanager/sync/model/OverwriteCloudShadowRequest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import java.time.Instant;
+import java.util.Optional;
+
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_LOCAL_VERSION_KEY;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_SHADOW_NAME_KEY;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_THING_NAME_KEY;
+
+public class OverwriteCloudShadowRequest extends BaseSyncRequest {
```
rbattle, line 23: does the RequestMerger need to be updated to support these new request types?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -211,130 +214,225 @@ void handleDeviceThingNameChange(WhatHappened whatHappened, Node changedNode) {
...
     @Override
     protected void install() {
+        install(InstallConfig.builder()
+                .configureMaxDocSizeLimitConfig(true)
+                .configureRateLimitsConfig(true)
+                .configureStrategyConfig(true)
+                .configureSyncDirectionConfig(true)
+                .configureSynchronizeConfig(true)
+                .installDatabase(true)
+                .build());
```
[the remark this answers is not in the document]
nikkhilmuthye: Unit tests should just test a unit of code right? I needed to write too many mocks and returns to test a small part of the code. So adding this helped reduce that a bit.

src/main/java/com/aws/greengrass/shadowmanager/sync/model/Direction.java
FThompsonAWS: Can we include the source and target in the direction to make it clear exactly what it does?
* DEVICE_TO_CLOUD
* CLOUD_TO_DEVICE

These names also are future proof if we ever support syncing between additional source/target types, such as other core devices (i.e. sources and targets other than the local device and cloud).

With future proofing in mind, should BIDRECTIONAL include the source and target as well? If so, here's a few suggestions:
* BETWEEN_DEVICE_AND_CLOUD
* BIDIRECTIONAL_DEVICE_AND_CLOUD

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -80,6 +84,42 @@ public void stopSubscribing() {
...
+        unsubscribeForAllShadowsTopics(subscribedDeleteShadowTopics, this::handleDelete);
+    }
+
+    /**
+     * Unsubscribe from all the shadow topics.
+     *
+     * @param topics   topics to unsubscribe
+     * @param callback Callback function applied to shadow topic
+     */
+    public synchronized void unsubscribeForAllShadowsTopics(Set<String> topics, Consumer<MqttMessage> callback) {
```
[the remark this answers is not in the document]
nikkhilmuthye: We have separate topics for update and delete, so we need to do it twice.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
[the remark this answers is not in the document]
nikkhilmuthye: So the reason I set it to false, is because most of the times the function is called with one field set to true and the rest as false. So that is clear what part of the `install` function we are trying to test.
Are you suggesting I remove this and have all the mocks in tests?

src/main/java/com/aws/greengrass/shadowmanager/sync/model/OverwriteCloudShadowRequest.java
nikkhilmuthye: Good point. This change needs some more refactoring. I will do that in the next PR.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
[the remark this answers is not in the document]
nikkhilmuthye: Spoke offline. We will extract out the configuration parsing into a separate class with specific methods like `parseSyncDirection`. This will enable a better coding pattern. I will handle this in the next PR

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncDirectionalityTest.java
```diff
@@ -0,0 +1,243 @@
...
+                .resetRetryConfig(false)
+                .build());
+        shadowManager.startSyncingShadows(ShadowManager.StartSyncInfo.builder().build());
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
+
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        syncHandler.pushLocalUpdateSyncRequest(MOCK_THING_NAME_1, CLASSIC_SHADOW, JsonUtil.getPayloadBytes(cloudDocument));
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
+
+        TimeUnit.SECONDS.sleep(2);
```
rbattle, line 235: is this sleep necessary? even with the eventuallyEval's below?

rbattle (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/ShadowIPCTest.java
```diff
@@ -139,26 +151,37 @@
...
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        SyncStrategy realTimeSyncStrategy = syncHandler.getOverallSyncStrategy();
+        // set retry config to only try once so we can test failures earlier
         kernel.launch();
-        assertTrue(awaitIpcServiceLatch.await(10, TimeUnit.SECONDS));
+
+        assertTrue(shadowManagerRunning.await(30L, TimeUnit.SECONDS));
+        realTimeSyncStrategy.stop();
+
+        isSyncMocked.set(true);
```
youtuyy, line 182: Is isSyncMocked used?

youtuyy (approved):

PR #106, opened 2022-01-06
fix: check if update is necessary before adding sync request.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Do not add the sync request in the queue if the update is not necessary. So we need to check that before we add it to the queue.
Also added some new booleans for tests to ensure that the queue is actually empty. 

We should not have a sync request which does not need to be processed in the queue. This was causing issue in the periodic sync wherein a local update was getting converted to a full sync due to an unnecessary local update request after the last cloud shadow update.

Updated the tests.

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
@@ -53,8 +53,6 @@ import java.util.Collections;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 
 import static com.aws.greengrass.shadowmanager.model.Constants.SHADOW_DOCUMENT_METADATA;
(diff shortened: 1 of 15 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -112,9 +116,11 @@
 
     private Supplier<Optional<SyncInformation>> syncInfo;
     private Supplier<Optional<ShadowDocument>> localShadow;
+    private final Lock sequential = new ReentrantLock();
 
     @BeforeEach
     void setup() {
+        sequential.lock();
```
rbattle, line 123: why do you need a lock in setup / cleanup?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -95,6 +98,29 @@
...
+     * @return true if Shadow Manger is syncing.
+     */
+    public boolean isSyncing() {
+        return syncing.get();
+    }
+
+    /**
+     * Getter for executing boolean.
+     *
+     * @return true if Shadow Manger is executing.
```
rbattle, line 118: nit: Manager

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -95,6 +98,29 @@
...
 
+    /**
+     * Indicates whether a sync request is executing or not.
+     */
+    AtomicBoolean executing = new AtomicBoolean(false);
+
+    /**
+     * Getter for syncing boolean.
+     *
+     * @return true if Shadow Manger is syncing.
```
rbattle, line 109: nit: Manager

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
nikkhilmuthye: This was one of the options I had tried to make the tests less flaky. I thought I had removed it. But removed it now.

rbattle (approved):

vabhasin (approved):

PR #110, opened 2022-01-13
fix: update sync version if update is deemed not necessary

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Update the version of the shadow if the update is deemed unnecessary for either a cloud update or a local update.

There might be times when a cloud shadow gets updated more than once without any change to the shadow contents. This causes the cloud shadow version to increment. Since we don't sync that version correctly, it causes a full sync on the next attempt.

Added unit tests.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
@@ -62,6 +62,7 @@ public class CloudUpdateSyncRequest extends BaseSyncRequest {
 
     /**
      * Merge the sync requests together.
+     *
      * @param other the newer request to merge
      */
     public void merge(CloudUpdateSyncRequest other) {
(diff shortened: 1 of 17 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
```diff
@@ -109,6 +109,7 @@ public void execute(SyncContext context) throws RetryableException, SkipSyncRequ
                     .kv(LOG_THING_NAME_KEY, getThingName())
                     .kv(LOG_SHADOW_NAME_KEY, getShadowName())
                     .kv(LOG_LOCAL_VERSION_KEY, currentSyncInformation.getLocalVersion())
+                    .kv(LOG_CLOUD_VERSION_KEY, currentSyncInformation.getCloudVersion())
                     .kv(LOG_CLOUD_VERSION_KEY, cloudUpdatedVersion)
```
rbattle, line 113: should you have a different key here?

rbattle (approved):

vabhasin (approved):

PR #112, opened 2022-01-26
fix: handle `AbortedException` correctly when syncing with cloud.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
If an `AbortedException` is thrown, check the cause to see if it was caused due to an `InterruptedException`. If it is, then rethrow that back to the caller. The `syncLoop` should add this request back in the queue if the sync request failed due to an `InterruptedException`.

There might be times when a sync request fails due to the thread being interrupted (while syncing) or if the thread just dies.

Added / Updated unit tests.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
@@ -91,7 +91,7 @@ public class PeriodicSyncStrategy extends BaseSyncStrategy {
      * Take and execute items from the sync queue. This is intended to be run in a separate thread.
      * This will work on all the sync requests in the queue until it's empty.
      */
-    @SuppressWarnings({"PMD.CompareObjectsWithEquals", "PMD.AvoidCatchingGenericException"})
+    @SuppressWarnings({"PMD.CompareObjectsWithEquals", "PMD.AvoidCatchingGenericException", "PMD.NullAssignment"})
     private void syncLoop() {
         // If another thread is already running, then return.
         if (!isRunning.compareAndSet(false, true)) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/PeriodicSyncStrategy.java
```diff
@@ -139,6 +139,8 @@ private void syncLoop() {
                     }
                 } catch (InterruptedException e) {
                     logger.atWarn(SYNC_EVENT_TYPE).log("Interrupted while waiting for sync requests");
+                    // Add the sync request back in the queue since we still need to process this later.
+                    syncQueue.offer(request);
```
rbattle, line 143: Is there still a chance that the thread is interrupted but no `InterruptedException` is thrown? We check the `Thread.currentThread().isInterrupted()`  in the while loop here, and in the do / while loop in the RealtimeSync.

Would it be better to set the request to null after processing (at line 119) and before picking up the next one. Then outside the loop, before the`finally` at line 172 you could check:
```
if (request != null) {
  syncQueue.offer(request);
}
```

rbattle (commented):
looks good - one small comment to consider

vabhasin (dismissed):

rbattle (dismissed):

rbattle:
not sure why lgtm thinks the variable could be `null` - the request is coming from `SyncRequest request = syncQueue.take();` and it's not null there.

MikeDombo:
> not sure why lgtm thinks the variable could be `null` - the request is coming from `SyncRequest request = syncQueue.take();` and it's not null there.

take returns the head, if the queue is empty then it would return null most likely. So yes, there should be a null check

nikkhilmuthye:
> > not sure why lgtm thinks the variable could be `null` - the request is coming from `SyncRequest request = syncQueue.take();` and it's not null there.
> 
> take returns the head, if the queue is empty then it would return null most likely. So yes, there should be a null check

take waits if the queue is empty. If another threads adds an item to the queue, this thread will get signed. So that time the head won't be null.

vabhasin (approved):

rbattle:
> > > not sure why lgtm thinks the variable could be `null` - the request is coming from `SyncRequest request = syncQueue.take();` and it's not null there.
> > 
> > 
> > take returns the head, if the queue is empty then it would return null most likely. So yes, there should be a null check
> 
> take waits if the queue is empty. If another threads adds an item to the queue, this thread will get signed. So that time the head won't be null.

you also can't add null objects to the queue

rbattle (approved):

PR #113, opened 2022-01-28
fix: wait for executing sync request to finish before cancelling threads

rbattle:
**Issue #, if available:**

**Description of changes:**
Cancelling sync threads can cause database exceptions if an I/O operation is occurring when the thread is cancelled.

A semaphore is used along with the existing  flag to force the stopper to wait until the request is finished.

SyncStrategy common logic was refactored into the BaseSyncStrategy class to avoid duplication. 

The `syncLoop` method from the `PeriodicSyncStrategy` class was pulled up to the `BaseSyncStrategy`.  Periodic still uses the non-blocking `poll` and RealTime still uses the blocking `take` to get data off the queue. A permit on the semaphore is acquired before executing the request and there are checks for whether syncing has stopped. A latch is used to wait for threads to exit.

Extra test cases have been added to cover this change, and an integration test that turns syncing on and off repeatedly for 30 seconds has been added.

To avoid H2 database exceptions and other Interrupted exceptions when syncing gets stopped while a request is processing.

Unit/integration

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
@@ -51,8 +51,14 @@ import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
@@ -69,6 +75,7 @@ import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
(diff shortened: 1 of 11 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -191,6 +192,40 @@ void GIVEN_sync_config_and_no_local_WHEN_startup_THEN_local_version_updated_via_
...
+
+            ShadowManager.StartSyncInfo syncInfo =
+                    ShadowManager.StartSyncInfo.builder().reInitializeSyncInfo(false).overrideRunningCheck(true).updateCloudSubscriptions(false).build();
+
+            SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+            Instant end = Instant.now().plusSeconds(30);
+            // this is more of a sanity check to ensure that cancelling threads doesn't cause any errors
+
+            // stop and start syncing for 30 seconds. Sync requests will be added when syncing starts.
+            while (end.isAfter(Instant.now())) {
```
nikkhilmuthye, line 219: should we have shadow operations on a separate thread happening here?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -103,6 +108,10 @@
      */
     AtomicBoolean executing = new AtomicBoolean(false);
 
+    Semaphore exec;
```
nikkhilmuthye, line 111: nit: Can we rename this to something that makes it more readable?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -103,6 +108,10 @@
...
+    int syncParallelism;
+    CountDownLatch latch;
```
nikkhilmuthye, line 113: nit: Same here. Can we rename?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -198,10 +219,51 @@ public void stop() {
...
+            // cancelling the sync threads.
+
+            // There is a semaphore around the critical section in the sync loop - we do not want to cancel while
+            // that is held.
+            // There is a latch that is decremented when the syncLoop exits - this allows us to know when stopped or
+            // cancelled threads have actually exited. Cancelling a future will not let you know when the thread has
+            // actually finished.
+            if (!exec.tryAcquire(syncParallelism)) {
+                logger.atInfo(SYNC_EVENT_TYPE).log("Waiting for in-flight sync requests to finish");
+                exec.acquireUninterruptibly(syncParallelism);
```
nikkhilmuthye, line 242: Why not just call this instead of `tryAcquire` and then `acquireUninterruptibly`?

rbattle: I wanted to log that it's going to wait - if I just called `acquireUniterruptibly` there's no opportunity to print a message

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
rbattle: they are happening on a separate thread - when syncing starts is uses the ExecutorService to start syncing on another thread

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
rbattle: renamed to `criticalExecBlock`

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
rbattle: renamed to `syncThreadEnd`

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
rbattle: it's got some full sync operations that occur already

nikkhilmuthye: Ok right. I forgot that even the Full sync request would get the sync info and shadow on even if there is nothing to really update. So that should touch the DB.

nikkhilmuthye (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
rbattle: I thought about it some more and added a thread that does random updates, as well as sleeping for random times to allow syncing to occur for various lengths of time

nikkhilmuthye (approved):

randolphtan121 (approved):
Great refactor with the SyncStrategyTestBase class

PR #114, opened 2022-02-02
fix: improvements to sync stopping behavior

rbattle:
**Issue #, if available:**

**Description of changes:**
Only restart syncing on config changes that make a change
Don't block while periodic sync threads are scheduled but not executing

Config can get updated frequently without changing (e.g. timestamp updated) and this causes unnecessary full syncs by restarting

Cancelling periodic sync doesn't need to wait when it's not executing

unit

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -302,9 +302,12 @@ public class ShadowManager extends PluginService {
                 });
 
         Topics strategyTopics = config.lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_STRATEGY_TOPIC);
+        final Strategy[] currentStrategy = { DEFAULT_STRATEGY };
+
         strategyTopics.subscribe((why, newv) -> {
             Strategy strategy;
             Map<String, Object> strategyPojo = strategyTopics.toPOJO();
+
             if (WhatHappened.removed.equals(why) || strategyPojo == null || strategyPojo.isEmpty()) {
                 strategy = DEFAULT_STRATEGY;
             } else {
@@ -316,9 +319,12 @@ public class ShadowManager extends PluginService {
                     return;
                 }
             }
-            stopSyncingShadows(false);
-            syncHandler.setSyncStrategy(strategy);
-            startSyncingShadows(StartSyncInfo.builder().build());
+            if (!currentStrategy[0].equals(strategy)) {
+                currentStrategy[0] = strategy;
+                stopSyncingShadows(false);
+                syncHandler.setSyncStrategy(strategy);
+                startSyncingShadows(StartSyncInfo.builder().build());
+            }
         });
     }
 
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -59,6 +59,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -302,9 +302,12 @@ protected void install() {
                 });
 
         Topics strategyTopics = config.lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_STRATEGY_TOPIC);
+        final Strategy[] currentStrategy = { DEFAULT_STRATEGY };
```
nikkhilmuthye, line 305: Why have an array? Can't we just an `AtomicReference`?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -316,9 +319,12 @@ protected void install() {
...
             }
-            stopSyncingShadows(false);
-            syncHandler.setSyncStrategy(strategy);
-            startSyncingShadows(StartSyncInfo.builder().build());
+            if (!currentStrategy[0].equals(strategy)) {
+                currentStrategy[0] = strategy;
+                stopSyncingShadows(false);
+                syncHandler.setSyncStrategy(strategy);
+                startSyncingShadows(StartSyncInfo.builder().build());
+            }
```
randolphtan121, line 327: Is this covered in a unit/integration test already?

src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -41,6 +39,7 @@
 public abstract class BaseSyncStrategy implements SyncStrategy {
     private static final Logger logger = LogManager.getLogger(BaseSyncStrategy.class);
     static final String SYNC_EVENT_TYPE = SYNC.code();
+    static final int THREAD_END_WAIT_TIME_SECONDS = 10;
```
randolphtan121, line 42: It looks like this changed from 30 seconds to a 10 second timeout, is there data supporting the quicker timeout?

rbattle: there's also no data supporting the longer one - 30 seconds was also an arbitrary number.

In the logs on one of the failed acceptance tests, syncing was stopped, and the IPC publisher thread was blocked for 30 seconds (due to it waiting for a latch that would never countdown)

I'd rather not block for 30 seconds - a single sync request shouldn't take 10s - except for an incredibly slow network

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
rbattle: The reference changes though because it creates a new strategy object, even though they are logically equal

so you can't use `compareAndSet` because it is doing `==`. Then it's no different - you have to do `!ar.get().equals(strategy)` and it's functionally no different.

What's wrong with the array approach?

nikkhilmuthye: I just think it looks a little cleaner rather than with an array.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -316,9 +319,12 @@ protected void install() {
                     return;
                 }
             }
-            stopSyncingShadows(false);
-            syncHandler.setSyncStrategy(strategy);
-            startSyncingShadows(StartSyncInfo.builder().build());
+            if (!currentStrategy[0].equals(strategy)) {
```
randolphtan121, line 322: Does this cover the other case where we are going from the Periodic (non-default) to RealTime (default) strategy? It looks like here that currentStrategy will always be (DEFAULT_STRATEGY) and if we convert back the code block doesn't get executed.

rbattle: whenever the strategy is updated we store a reference to it and then compare it to the new one

In this case, we start at DEFAULT if there is no config and store that. Then we update to periodic, and that replaces the value. Then we switch back to DEFAULT and that again replaces the stored value.

randolphtan121 (approved):

nikkhilmuthye (approved):

PR #115, opened 2022-02-07
feat: add support for shadow operations over local PubSub

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
This will add support for Shadow Manager operations over local PubSub. This change is contingent on the [Nucleus PubSub PR](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1148).

This would enable customers to perform shadow operations over local PubSub. This would also open doors for Client device integration.

Added new unit tests.
Integration tests will be added once the  [Nucleus PubSub PR](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1148) is merged in.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.shadowmanager;
+
+import com.aws.greengrass.builtin.services.pubsub.PublishEvent;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowRequestHandler;
+import com.aws.greengrass.shadowmanager.ipc.GetThingShadowRequestHandler;
+import com.aws.greengrass.shadowmanager.ipc.PubSubClientWrapper;
+import com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler;
+import com.aws.greengrass.shadowmanager.model.LogEvents;
+import com.aws.greengrass.shadowmanager.model.ShadowRequest;
+import software.amazon.awssdk.aws.greengrass.model.DeleteThingShadowRequest;
+import software.amazon.awssdk.aws.greengrass.model.GetThingShadowRequest;
+import software.amazon.awssdk.aws.greengrass.model.UpdateThingShadowRequest;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_OPERATION;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_SHADOW_NAME_KEY;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_THING_NAME_KEY;
+import static com.aws.greengrass.shadowmanager.model.Constants.LOG_TOPIC;
+import static com.aws.greengrass.shadowmanager.model.Constants.SHADOW_MANAGER_NAME;
+
+
+/**
+ * Class that handles PubSub shadow topic subscription and managing any publish events on the shadow topic.
+ */
+public class PubSubIntegrator {
+    private static final Logger logger = LogManager.getLogger(PubSubIntegrator.class);
+
+    private final DeleteThingShadowRequestHandler deleteThingShadowRequestHandler;
+    private final UpdateThingShadowRequestHandler updateThingShadowRequestHandler;
+    private final GetThingShadowRequestHandler getThingShadowRequestHandler;
+    private final PubSubClientWrapper pubSubClientWrapper;
+    private final Pattern shadowPattern = Pattern.compile("\\$aws\\/things\\/(.*)\\/shadow(\\/name\\/(.*))?"
+            + "\\/(update|delete|get)");
+    private final AtomicBoolean subscribed = new AtomicBoolean(false);
+
+    /**
+     * Constructor for PubSubIntegrator.
+     *
+     * @param pubSubClientWrapper             The PubSub client wrapper
+     * @param deleteThingShadowRequestHandler handler class to handle the Delete Shadow request.
+     * @param updateThingShadowRequestHandler handler class to handle the Update Shadow request.
+     * @param getThingShadowRequestHandler    handler class to handle the Get Shadow request.
+     */
+    public PubSubIntegrator(PubSubClientWrapper pubSubClientWrapper,
+                            DeleteThingShadowRequestHandler deleteThingShadowRequestHandler,
+                            UpdateThingShadowRequestHandler updateThingShadowRequestHandler,
+                            GetThingShadowRequestHandler getThingShadowRequestHandler) {
+        this.pubSubClientWrapper = pubSubClientWrapper;
+        this.deleteThingShadowRequestHandler = deleteThingShadowRequestHandler;
+        this.updateThingShadowRequestHandler = updateThingShadowRequestHandler;
+        this.getThingShadowRequestHandler = getThingShadowRequestHandler;
+    }
+
+    /**
+     * Subscribes to the shadow topic over local PubSub.
+     */
+    public void subscribe() {
+        if (this.subscribed.compareAndSet(false, true)) {
+            this.pubSubClientWrapper.subscribe(this::handlePublishedMessage);
+        }
+    }
+
+    /**
+     * Unsubscribes to the shadow topic over local PubSub.
+     */
+    public void unsubscribe() {
+        if (this.subscribed.compareAndSet(true, false)) {
+            this.pubSubClientWrapper.unsubscribe(this::handlePublishedMessage);
+        }
+
+    }
+
+    /**
+     * Helper function to extract the thingName, shadowName and shadow operation from mqtt topic.
+     *
+     * @param topic PubSub topic on which the message was sent.
+     * @return ShadowRequest object with shadow details
+     */
+    ShadowRequest extractShadowFromTopic(String topic) {
+        final Matcher matcher = shadowPattern.matcher(topic);
+
+        if (matcher.find() && matcher.groupCount() == 4) {
+            String thingName = matcher.group(1);
+            String shadowName = matcher.group(3);
+            String operation = matcher.group(4);
+            return new ShadowRequest(thingName, shadowName, operation);
+        }
+        logger.atWarn()
+                .kv("topic", topic)
+                .log("Unable to parse shadow topic for thing name, shadow name and shadow operation");
+        throw new IllegalArgumentException("Unable to parse shadow topic for thing name shadow name "
+                + "and shadow operation");
+    }
+
+    /**
+     * Handle the new message that is published oveg local PubSub. Extract the necessary information from the shadow
+     * topic to accurately route the message to the appropriate handler.
+     * It will ignore the message if it is unable to extract all the necessary information from the shadow topic.
+     *
+     * @param publishEvent The message that is published over local PubSub.
+     */
+    private void handlePublishedMessage(PublishEvent publishEvent) {
+        String topic = publishEvent.getTopic();
+        logger.atDebug().kv(LOG_TOPIC, topic).log("Processing new shadow operation message over local PubSub");
+        try {
+            ShadowRequest shadowRequest = extractShadowFromTopic(topic);
+            switch (shadowRequest.getOperation().toLowerCase()) {
+                case "update":
+                    UpdateThingShadowRequest request = new UpdateThingShadowRequest();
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Handle the new message that is published oveg local PubSub. Extract the necessary information from the shadow
+     * topic to accurately route the message to the appropriate handler.
+     * It will ignore the message if it is unable to extract all the necessary information from the shadow topic.
+     *
+     * @param publishEvent The message that is published over local PubSub.
+     */
+    private void handlePublishedMessage(PublishEvent publishEvent) {
```
randolphtan121, line 112: Are we omitting the ListNamedShadowsForThing operation from local pubsub?

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+            String thingName = matcher.group(1);
+            String shadowName = matcher.group(3);
+            String operation = matcher.group(4);
+            return new ShadowRequest(thingName, shadowName, operation);
+        }
+        logger.atWarn()
+                .kv("topic", topic)
+                .log("Unable to parse shadow topic for thing name, shadow name and shadow operation");
+        throw new IllegalArgumentException("Unable to parse shadow topic for thing name shadow name "
+                + "and shadow operation");
```
rbattle, line 102: nit: "Unable to parse shadow topic" should be fine - otherwise add a comma between 'thing name, shadow name, and shadow operation'

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+                    logger.atWarn()
+                            .kv(LOG_THING_NAME_KEY, shadowRequest.getThingName())
+                            .kv(LOG_SHADOW_NAME_KEY, shadowRequest.getShadowName())
+                            .kv(LOG_OPERATION, shadowRequest.getOperation())
+                            .log("Unable to perform shadow operation due to unknown value");
+                    break;
+            }
+            logger.atDebug().kv(LOG_TOPIC, topic)
+                    .log("Finished processing new shadow operation message over local PubSub");
+        } catch (IllegalArgumentException e) {
```
rbattle, line 165: can this be thrown from any where in the switch? If this is the only exception you are catching and it can only come from `extractShadowFromTopic` then put the exception handling nearer the source of the exception:
```
ShadowRequest shadowrequest = null;
try {
  shadowRequest = extractShadowFromTopic();
} catch (IllegalArgumentException e) {
  return;
}
switch () {
 ...
}
```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -406,6 +410,7 @@ protected void shutdown() throws InterruptedException {
             stopSyncingShadows(true);
             database.close();
             inboundRateLimiter.clear();
+            pubSubIntegrator.unsubscribe();
```
rbattle, line 413: looking at this order of shutdown, can you switch it to:
```
stopSyncingShadows(true);
pubSubIntegrator.unsubscribe();
inboundRateLimiter.clear();
database.close();
```

It seems like we should first stop things that generate work for us (syncing, handling IPC subscriptions), then clear the rate limiter, and finally stop the database.

We don't want to close the database while we are still able to receive messages.

Also, if we are in the middle of processing an IPC message and we get shutdown, do we do anything to "finish" the IPC request?

src/main/java/com/aws/greengrass/shadowmanager/ipc/PubSubClientWrapper.java
```diff
@@ -113,4 +117,22 @@ private String getShadowPublishTopic(PubSubRequest ipcRequest, String topic) {
...
+     * @param cb Cosnumer to invoke upon receiving a new message over local PubSub.
+     */
+    public void subscribe(Consumer<PublishEvent> cb) {
+        this.pubSubIPCEventStreamAgent.subscribe(PUBSUB_SUBSCRIBE_TOPIC, cb, SHADOW_MANAGER_NAME);
+    }
+
+    /**
+     * Unsubscribes to the shadow topic over local PubSub.
+     *
+     * @param cb Cosnumer to invoke upon receiving a new message over local PubSub.
```
rbattle, line 133: nit: Consumer

src/main/java/com/aws/greengrass/shadowmanager/model/ShadowRequest.java
```diff
@@ -21,11 +21,12 @@
 public class ShadowRequest {
     String thingName;
     String shadowName;
+    String operation;
```
rbattle, line 24: should this be an enum to avoid the .toLowerCase()? The constructor could still take a string

src/test/java/com/aws/greengrass/shadowmanager/PubSubIntegratorTest.java
```diff
@@ -0,0 +1,181 @@
...
+        integrator.unsubscribe();
+        verify(mockPubSubClientWrapper, atMostOnce()).unsubscribe(any());
+
+        integrator.unsubscribe();
+        verify(mockPubSubClientWrapper, atMostOnce()).unsubscribe(any());
+    }
+
+    @ParameterizedTest
+    @MethodSource("classicAndNamedShadow")
+    void GIVEN_classic_shadow_op_invocation_WHEN_accept_THEN_calls_the_corret_handler(String shadowName, String op, ExtensionContext extensionContext) {
```
rbattle, line 117: nit: correct_handler

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+                            .log("Successfully deleted shadow");
+                    break;
+                case "get":
+                    GetThingShadowRequest getRequest = new GetThingShadowRequest();
+                    getRequest.setThingName(shadowRequest.getThingName());
+                    getRequest.setShadowName(shadowRequest.getShadowName());
+
+                    this.getThingShadowRequestHandler.handleRequest(getRequest, SHADOW_MANAGER_NAME);
+                    logger.atTrace()
+                            .setEventType(LogEvents.DELETE_THING_SHADOW.code())
```
rbattle, line 150: nit: GET

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+                    request.setThingName(shadowRequest.getThingName());
+                    request.setShadowName(shadowRequest.getShadowName());
+                    request.setPayload(publishEvent.getPayload());
+
+                    this.updateThingShadowRequestHandler.handleRequest(request, SHADOW_MANAGER_NAME);
+                    logger.atTrace()
+                            .setEventType(LogEvents.UPDATE_THING_SHADOW.code())
+                            .kv(LOG_THING_NAME_KEY, shadowRequest.getThingName())
+                            .kv(LOG_SHADOW_NAME_KEY, shadowRequest.getShadowName())
+                            .log("Successfully updated shadow");
```
rbattle, line 129: Do you need these logs? The handlers themselves [log](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/main/src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java#L227)

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,173 @@
...
+        try {
+            ShadowRequest shadowRequest = extractShadowFromTopic(topic);
+            switch (shadowRequest.getOperation().toLowerCase()) {
+                case "update":
+                    UpdateThingShadowRequest request = new UpdateThingShadowRequest();
+                    request.setThingName(shadowRequest.getThingName());
+                    request.setShadowName(shadowRequest.getShadowName());
+                    request.setPayload(publishEvent.getPayload());
+
+                    this.updateThingShadowRequestHandler.handleRequest(request, SHADOW_MANAGER_NAME);
```
rbattle, line 124: what happens on a ServiceError - does that bubble up and just get logged?

Can you add some test cases for the handler throwing an exception?

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
randolphtan121, line 112: We should add the `ListNamedShadowForThing` request as well here as it is a supported shadow command

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
randolphtan121: +1 for "Unable to parse shadow topic". Operation is optional and may lead to confusion if user did not mean to supply an operation.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
nikkhilmuthye: Right now, we don't do anything. It will just stop. Do we want to add an `AtomicInteger` to keep track of how many requests are being processed right now and only shut down if and when the count becomes 0?

src/main/java/com/aws/greengrass/shadowmanager/model/ShadowRequest.java
nikkhilmuthye: We can make it into an enum. Might look cleaner. I just thought for one `toLowerCase` function, do we need to add an enum?

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
nikkhilmuthye: True. I just thought it might be a good data point for the customer to know if this operation happened over IPC PubSub or IPC shadow. I can add a kv specific to pubsub to these logs. Hence I put it in trace too.

rbattle: but they would see the log on line 114 on the same thread that it is "Processing new shadow operation message over local pubsub" with the topic - which would include the operation right? 

It would look like:

```
[date] [DEBUG] (<thread-id-1>) c.a.g.sm.PubSubIntegrator: Processing new shadow operation message over local PubSub. {topic=$aws/things/<thingName>/shadow/<shadowName>/update}
[date] [INFO] (<thread-id-1>) c.a.g.sm.ipc.UpdateThingShadowRequestHandler: Successfully updated shadow. {thing name=<thingName>, shadow name=<shadowName>, local-version=<version>, service-name=<serviceName>}
[date] [TRACE] (<thread-id-1>) c.a.g.sm.PubSubIntegrator: handle-update-thing-shadow. Successfully updated shadow. {thing name=<thingName>, shadow name=<shadow Name>
```

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,172 @@
...
+            return new ShadowRequest(thingName, shadowName, operation);
+        }
+        logger.atWarn()
+                .kv("topic", topic)
+                .log("Unable to parse shadow topic for thing name, shadow name and shadow operation");
+        throw new IllegalArgumentException("Unable to parse shadow topic");
+    }
+
+    /**
+     * Handle the new message that is published oveg local PubSub. Extract the necessary information from the shadow
```
youtuyy, line 107: typo: over

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
nikkhilmuthye: Spoke offline. Decided to punt this for now.

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    private void handlePublishedMessage(PublishEvent publishEvent) {
+        String topic = publishEvent.getTopic();
+        logger.atDebug().kv(LOG_TOPIC, topic).log("Processing new shadow operation message over local PubSub");
+        ShadowRequest shadowRequest = null;
+        try {
+            shadowRequest = extractShadowFromTopic(topic);
+        } catch (IllegalArgumentException e) {
+            logger.atWarn()
+                    .setEventType(LogEvents.UPDATE_THING_SHADOW.code())
```
youtuyy, line 121: do you need the event type as UPDATE_THING_SHADOW ?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
rbattle: I think that ideally we could finish in flight requests. For syncing it will tell it to cancel the thread, but if a request is executing it will wait for that request to finish.

Let's not add extra complexity to this PR though

rbattle (approved):

youtuyy (approved):

PR #117, opened 2022-02-23
fix: ignore shadow response messages.

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Ignore response messages for shadow operations over PubSub.

We do not need to process the shadow response messages in Shadow Manager.

Added unit test

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
@@ -42,6 +42,8 @@ public class PubSubIntegrator {
     private final PubSubClientWrapper pubSubClientWrapper;
     private final Pattern shadowPattern = Pattern.compile("\\$aws\\/things\\/(.*)\\/shadow(\\/name\\/(.*))?"
             + "\\/(update|delete|get)");
+    private final Pattern shadowResponsePattern = Pattern.compile("\\$aws\\/things\\/(.*)\\/shadow(\\/name\\/(.*))?"
+            + "\\/(update|delete|get)\\/(accepted|rejected|delta|documents)");
     private final AtomicBoolean subscribed = new AtomicBoolean(false);
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -111,6 +113,13 @@ ShadowRequest extractShadowFromTopic(String topic) {
      */
     private void handlePublishedMessage(PublishEvent publishEvent) {
         String topic = publishEvent.getTopic();
+
+        if (isResponseMessage(topic)) {
+            logger.atDebug().kv(LOG_TOPIC, topic)
+                    .log("Not processing message since it is a response message to a shadow operation");
+            return;
```
youtuyy, line 120: Will it work that we don't subscribe to those response topics at all but only subscribe to the ones we need to react to?

nikkhilmuthye: We had a broader discussion around this today. We might need some work in PubSub.

vabhasin: > Will it work that we don't subscribe to those response topics at all but only subscribe to the ones we need to react to?

+1

src/main/java/com/aws/greengrass/shadowmanager/PubSubIntegrator.java
```diff
@@ -167,4 +176,15 @@ private void handlePublishedMessage(PublishEvent publishEvent) {
...
+
+    /**
+     * Checks if the message received is a response message or not.
+     *
+     * @param topic the topic on which the PubSub message was sent.
+     * @return true if the message is a response message to a shadow operation; Else false.
+     */
+    private boolean isResponseMessage(String topic) {
+        final Matcher matcher = shadowResponsePattern.matcher(topic);
+        return matcher.find() && matcher.groupCount() == 5;
```
rbattle, line 188: why check the group count? This just tells you how many groups are in the pattern, not whether they all matched.

e.g.:
```
$aws/things/foo/shadow/name/bar/update/rejected
``` 
has values for all 5 groups

```
$aws/things/foo/shadow/update/rejected
```
has null values for groups 2 and 3

Is it enough to just do `return matcher.matches()`?

youtuyy (approved):

rbattle (approved):

PR #119, opened 2022-03-10
feat: add sync direction in `RequestMerger` to merge appropriate sync requests

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**

Add ability to set the `syncDirection` in the `RequestMerger` class from the `ShadowManager` configuration subscription.

Also inject the `RequestMerger` class instead of creating a new one every time.

`RequestMerger` needs to know the exact direction in which the sync is configured in order to properly merge in new sync requests based on it.

Added unit tests.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -33,7 +33,6 @@ import com.aws.greengrass.shadowmanager.model.configuration.ThingShadowSyncConfi
 import com.aws.greengrass.shadowmanager.model.dao.SyncInformation;
 import com.aws.greengrass.shadowmanager.sync.CloudDataClient;
 import com.aws.greengrass.shadowmanager.sync.IotDataPlaneClientWrapper;
-import com.aws.greengrass.shadowmanager.sync.RequestMerger;
 import com.aws.greengrass.shadowmanager.sync.SyncHandler;
 import com.aws.greengrass.shadowmanager.sync.model.Direction;
 import com.aws.greengrass.shadowmanager.sync.model.SyncContext;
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/RequestMerger.java
```diff
@@ -27,6 +32,9 @@
 public class RequestMerger {
     private static final Logger logger = LogManager.getLogger(RequestMerger.class);
 
+    @Setter
+    private Direction syncDirection = Direction.BETWEEN_DEVICE_AND_CLOUD;
```
youtuyy, line 36: Does it mean it's the default of the sync direction? Is there any default initializer for the Direction class we can use instead of setting the value here?

youtuyy (approved):

src/main/java/com/aws/greengrass/shadowmanager/sync/RequestMerger.java
nikkhilmuthye: I had to add this here for unit tests. For some reason it was set to null for some reason if it was not set. It is not the case in integ tests since we will always set it when we do the subscriptions. We also have a default value in Shadow Manager recipe which should set it.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -356,6 +361,7 @@ protected void install(InstallConfig installConfig) {
 
                         setupSync(newSyncDirection, Optional.of(currentDirection));
                         syncHandler.setSyncDirection(newSyncDirection);
+                        merger.setSyncDirection(newSyncDirection);
```
rbattle, line 364: could you have the RequestMerger in the `SyncHandler` so you could set the direction for the merger in `syncHandler.setSyncDirection`?

youtuyy (approved):

rbattle (approved):

Issue #126, opened 2022-09-29
(shadowmanager): Reports payload exceeds maximum size allowed when it doesnt


crispinboylan-int:
**Describe the bug**
We have 30k shadows, and shadow manager is updated with the maxShadowDocumentSizeBytes (30720).

When we have a large shadow close to the 30k limit we receive this error in shadow manager, and we cant find a way to get the shadow successfully synced again even if we remove data because it cant process that version.

If the cloud shadow api allows the update to the shadow, why does shadow manager not?

**To Reproduce**

Have a shadow close to the limit size.

**Expected behavior**
Accepted the update, as per the ap.

**Actual behavior**

Receive the following log message:

2022-09-29T15:40:39.487Z [INFO] (pool-2-thread-34) com.aws.greengrass.shadowmanager.sync.strategy.BaseSyncStrategy: sync. Executing sync request. {Type=LocalUpdateSyncRequest, thing name=home-OUnWKRu7Pn1B, shadow name=RoomConfig}
2022-09-29T15:40:39.505Z [WARN] (pool-2-thread-34) com.aws.greengrass.shadowmanager.ipc.BaseRequestHandler: handle-update-thing-shadow. {thing name=home-OUnWKRu7Pn1B, shadow name=RoomConfig}
com.aws.greengrass.shadowmanager.exception.InvalidRequestParametersException: The payload exceeds the maximum size allowed
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler.lambda$handleRequest$2(UpdateThingShadowRequestHandler.java:137)
        at com.aws.greengrass.ipc.common.ExceptionUtil.translateExceptions(ExceptionUtil.java:33)
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler.handleRequest(UpdateThingShadowRequestHandler.java:98)
        at com.aws.greengrass.shadowmanager.sync.model.LocalUpdateSyncRequest.execute(LocalUpdateSyncRequest.java:121)
        at com.aws.greengrass.shadowmanager.sync.SyncHandler.lambda$new$0(SyncHandler.java:113)
        at com.aws.greengrass.util.RetryUtils.runWithRetry(RetryUtils.java:50)
        at com.aws.greengrass.shadowmanager.sync.SyncHandler.lambda$new$1(SyncHandler.java:111)
[...]

PR #127, opened 2022-09-30
fix: validate shadow size after merging update

rbattle:
Shadow size validation needs to occur after merging the update request into the existing document.

Shadow size validation should just based on desired and reported nodes in order to be consistent with cloud.

**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-shadow-manager/issues/126

**Description of changes:**
Validate shadow size in a manner consistent with cloud

Extra bytes in shadow update are counted when it should be just the contents of "desired" and "reported"
Customers could potentially have larger shadows by setting multiple keys as 8kb in distinct requests, it needs to check shadow size after applying the update.

Unit tests. Ideally we could replicate with customer issue as well. Posting here for review

**Checklist:**
 

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -321,7 +321,6 @@ protected void install(InstallConfig installConfig) {
                         try {
```
MikeDombo, line 321: ^^
For all these configs, can we just do a read-through instead of copying data into fields when there is a change?

Much simpler and more correct if we can just read the currently configured value (or default) when it is needed. Then there's no possibility of a discrepancy between what we're using for the value and what was configured by the user.

Different PR though.

src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
```diff
@@ -178,17 +166,36 @@ public UpdateThingShadowHandlerResponse handleRequest(UpdateThingShadowRequest r
...
+                    if (!isNullOrMissing(updatedDocument.getState().getReported())) {
+                        reportedLength = JsonUtil.getPayloadBytes(updatedDocument.getState().getReported()).length;
+                    }
+
+                    // Make sure new document is not too big
+                    logger.atDebug()
+                            .setEventType(LogEvents.UPDATE_THING_SHADOW.code())
+                            .kv(LOG_THING_NAME_KEY, thingName)
+                            .kv(LOG_SHADOW_NAME_KEY, shadowName)
+                            .kv("mergedShadowStateSize", desiredLength + reportedLength)
```
MikeDombo, line 188: log what our max allowed size is

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
```diff
@@ -235,6 +242,8 @@ public UpdateThingShadowHandlerResponse handleRequest(UpdateThingShadowRequest r
                     this.syncHandler.pushCloudUpdateSyncRequest(thingName, shadowName, updateDocumentRequest);
 
                     return new UpdateThingShadowHandlerResponse(updateThingShadowResponse, updateDocumentBytes);
+                } catch (InvalidRequestParametersException e) {
```
nikkhilmuthye, line 245: we want throw back the correct exception message here right?  `ErrorMessage.PAYLOAD_TOO_LARGE_MESSAGE`

src/main/java/com/aws/greengrass/shadowmanager/util/Validator.java
```diff
@@ -150,4 +148,15 @@ public static void validateLocalShadowRequestsPerThingPerSecond(int maxLocalShad
         }
     }
 
+    /**
+     * Validate that a shadow size does not exceed the maximum.
+     * @param size size of shadow
+     * @throws InvalidRequestParametersException if the shadow is too large
+     */
+    public static void validateShadowSize(int size) throws InvalidRequestParametersException {
+        if (size > Validator.getMaxShadowDocumentSize()) {
```
nikkhilmuthye, line 157: nit: why are we using the getter here and not the variable itself?

nikkhilmuthye (approved):

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
rbattle: yeah - wanted that as a separate PR

src/main/java/com/aws/greengrass/shadowmanager/util/Validator.java
rbattle: copy/paste error

src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
rbattle: added a log when it loads the config

rbattle: and added an extra warning log when it encounters the problem so it can log that detail

MikeDombo (commented):
Validated this is working using the example provided in #126?
https://github.com/aws-greengrass/aws-greengrass-shadow-manager/issues/126#issuecomment-1279743773

MikeDombo (approved):

saranyailla (approved):

jcosentino11:
> Validated this is working using the example provided in #126? [#126 (comment)](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/issues/126#issuecomment-1279743773)

I wasn't able to use that specific data yet (waiting on limit increase in the account I was testing with), but was able to reproduce the issue manually using generated data.  Confirmed manually that this PR addresses the issue.

Issue #129, opened 2022-10-26
(configuration): ShadowManager stops responding to configuration updates


jbutler:
**Describe the bug**
Shadow Manager can reach a state where it effectively stops listening to configuration updates. While in this state, it is not possible for customers to update ShadowManager configuration to add / remove shadows, or to change sync behavior. This leads to customer frustration since their configuration may actually be correct, but ShadowManager does not respect it. Restarting Greengrass resolves the issue.

This issue is caused by the fact that ShadowManager subscribes to individual configuration nodes during initialization (e.g. [here](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/5a16d1723377a41aee6d270cffc3dbd929ef49f4/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L245), and [here](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/5a16d1723377a41aee6d270cffc3dbd929ef49f4/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L281)). These configuration node subscriptions are "lost" if the node gets deleted. This occurs if a deployment which performs a `RESET` on the configuration node occurs.

**To Reproduce**
1. Deploy ShadowManager component
2. Perform a second deployment which performs a ShadowManager configuration reset (`RESET: [""]`)
3. Perform a third deployment which configures SM to sync a shadow

**Expected behavior**
ShadowManager should respond to the 3rd deployment by beginning to sync the specified shadow.

**Actual behavior**
It does not sync the new shadow until Greengrass is restarted.

**Environment**
 - OS: [e.g. Ubuntu 20.04]
 - JDK version:
 - Nucleus version:
 - Shadow Manager version:

**Additional context**

E.g. what is the impact of the bug?

saranyailla:
Closed as part of [shadow manager 2.2.4 release]( https://github.com/aws-greengrass/aws-greengrass-shadow-manager/releases/tag/v2.2.4).

PR #131, opened 2022-11-09
fix: respond to config updates after config reset

saranyailla:
**Issue #, if available:**
#129 

**Description of changes:**
Subscribe to the changes on the parent configuration node instead of specific child nodes as the child nodes can be removed during reset. 

If the configuration is updated after reset, the component should behave according to the configuration changes without restarting the component. Without this change, the configuration won't be updated once the reset happens. 

Updated tests. Need to do manual testing. 

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -13,6 +13,8 @@ import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
 import com.aws.greengrass.shadowmanager.exception.SkipSyncRequestException;
 import com.aws.greengrass.shadowmanager.model.LogEvents;
 import com.aws.greengrass.shadowmanager.model.dao.SyncInformation;
+import com.aws.greengrass.shadowmanager.sync.SyncHandler;
+import com.aws.greengrass.shadowmanager.sync.model.Direction;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
 import com.aws.greengrass.util.Pair;
 import org.flywaydb.core.api.FlywayException;
@@ -38,10 +40,13 @@ import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_CLA
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_NAMED_SHADOWS_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_SHADOW_DOCUMENTS_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_SYNCHRONIZATION_TOPIC;
+import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_SYNC_DIRECTION_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_THING_NAME_TOPIC;
+
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -126,6 +131,7 @@ class ShadowManagerTest extends NucleusLaunchUtils {
         MqttClient mqttClient = mock(MqttClient.class);
         lenient().when(mqttClient.connected()).thenReturn(false);
 
+
         kernel.getContext().put(MqttClient.class, mqttClient);
         startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
                 .configFile(DEFAULT_CONFIG)
@@ -170,5 +176,35 @@ class ShadowManagerTest extends NucleusLaunchUtils {
                 new Pair<>(THING_NAME2, ""),
                 new Pair<>(THING_NAME2, "Shadow-0"),
                 new Pair<>(THING_NAME2, "Shadow-5")));
+    }
+
+
+    @Test
+    @SuppressWarnings("PMD.CloseResource")
+    void GIVEN_shadow_manager_WHEN_individual_config_resets_THEN_respond_to_config_updates(ExtensionContext context)
+            throws Exception {
+        ignoreExceptionOfType(context, SkipSyncRequestException.class);
+        MqttClient mqttClient = mock(MqttClient.class);
+        lenient().when(mqttClient.connected()).thenReturn(false);
+
+
+        kernel.getContext().put(MqttClient.class, mqttClient);
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile(DEFAULT_CONFIG)
+                .mqttConnected(false)
+                .build());
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        assertThat(syncHandler.getSyncDirection(), is(Direction.BETWEEN_DEVICE_AND_CLOUD));
+
+        shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC)
+                .lookup(CONFIGURATION_SYNC_DIRECTION_TOPIC).remove();
+        kernel.getContext().waitForPublishQueueToClear();
+        assertThat(syncHandler.getSyncDirection(), is(Direction.BETWEEN_DEVICE_AND_CLOUD));
+
+        shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC)
+                .lookup(CONFIGURATION_SYNC_DIRECTION_TOPIC).withValue(Direction.DEVICE_TO_CLOUD.getCode());
+        kernel.getContext().waitForPublishQueueToClear();
+
+        assertThat(syncHandler.getSyncDirection(), is(Direction.DEVICE_TO_CLOUD));
     }
 }
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -194,7 +194,10 @@ class ShadowManagerTest extends NucleusLaunchUtils {
                 .mqttConnected(false)
                 .build());
         SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
-        assertThat(syncHandler.getSyncDirection(), is(Direction.BETWEEN_DEVICE_AND_CLOUD));
+        shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC)
+                .lookup(CONFIGURATION_SYNC_DIRECTION_TOPIC).withValue(Direction.DEVICE_TO_CLOUD.getCode());
+        kernel.getContext().waitForPublishQueueToClear();
+        assertThat(syncHandler.getSyncDirection(), is(Direction.DEVICE_TO_CLOUD));
 
         shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC)
                 .lookup(CONFIGURATION_SYNC_DIRECTION_TOPIC).remove();
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
@@ -360,7 +360,7 @@ class SyncTest extends NucleusLaunchUtils {
                 .mockCloud(true)
                 .mockDao(true)
                 .build());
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
         }
 
         inboundRateLimiter.clear();
+        config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((what, newv) -> {
+            if (what.equals(WhatHappened.timestampUpdated)) {
+                return;
+            }
+            if (installConfig.configureSynchronizeConfig
+                    && (newv == null || newv.childOf(CONFIGURATION_SYNCHRONIZATION_TOPIC))) {
+                configureSynchronization();
```
MikeDombo, line 248: are any of these expensive in any way? I think I'd like to just run them all no matter what.

In fact, for the things which don't need to be event based (ie. they're just changing a config option like the max size) then I'd like them to not exist at all. I want the code to directly read from the config instead of reading a local field which is updated by the config (since that introduces an opportunity to be out of sync).

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
         }
 
         inboundRateLimiter.clear();
+        config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((what, newv) -> {
+            if (what.equals(WhatHappened.timestampUpdated)) {
+                return;
+            }
+            if (installConfig.configureSynchronizeConfig
```
Nelsonochoam, line 246: Why do we even have the `installConfig` seems like it just gets built on the `install` and all the values are set to true and it is not changing based on anything.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
rbattle: +1 - the config parsing/validation has gotten a bit too defensive - it could be simplified - e.g. if anything under syncing changes - let the sync handler go and get whatever config it needs and restart. If rate limits change, tell the rate limiter to reload values

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
...
+                    && (newv == null || newv.childOf(CONFIGURATION_SYNCHRONIZATION_TOPIC))) {
+                configureSynchronization();
+            }
+            if (installConfig.configureRateLimitsConfig
+                    && (newv == null || newv.childOf(CONFIGURATION_RATE_LIMITS_TOPIC))) {
+                rateLimits();
```
Nelsonochoam, line 252:
```suggestion
                configureRateLimits();
```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
...
+            }
+            if (installConfig.configureMaxDocSizeLimitConfig
+                    && (newv == null || newv.childOf(CONFIGURATION_MAX_DOC_SIZE_LIMIT_B_TOPIC))) {
+                maxSizeDocLimitConfig();
```
Nelsonochoam, line 256:
```suggestion
                configureMaxSizeDocLimit();
```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
...
+            }
+            if (installConfig.configureSyncDirectionConfig
+                    && (newv == null || newv.childOf(CONFIGURATION_SYNC_DIRECTION_TOPIC))) {
+                syncDirection();
```
Nelsonochoam, line 260:
```suggestion
                configureSynDirection();
```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
Nelsonochoam: Also is there a ways to not use any of this conditionals at all?. What if we just call `configureSynchronization` without any conditional and the we have the method read the configuration by doing a lookup

```java
Topics configTopics = config.lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC);
```

And decide what to do based on the value it currently holds?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,168 @@ protected void install(InstallConfig installConfig) {
...
+            }
+            if (installConfig.configureStrategyConfig
+                    && (newv == null || newv.childOf(CONFIGURATION_STRATEGY_TOPIC))) {
```
Nelsonochoam, line 263: Might be a dumb question but if the config previously had a value for `CONFIGURATION_STRATEGY_TOPIC` and then ONLY that key gets removed from the config would `newv == null`?. I see below we do a call for `toPojo` but would removing the key would cause it to return null or an empty map?

saranyailla: `newv` will be the topic that is removed - `services.aws.greengrass.ShadowManager.configuration.strategy` with its old value.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
saranyailla: Yes, we'll need address those in later PRs.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -171,4 +177,34 @@ void GIVEN_existing_sync_information_WHEN_config_updates_THEN_removed_sync_infor
...
+
+        kernel.getContext().put(MqttClient.class, mqttClient);
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile(DEFAULT_CONFIG)
+                .mqttConnected(false)
+                .build());
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        assertThat(syncHandler.getSyncDirection(), is(Direction.BETWEEN_DEVICE_AND_CLOUD));
+
+        shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_SYNCHRONIZATION_TOPIC)
```
Nelsonochoam, line 199: Mmm the `DEFAULT_CONFIG` doesn't specify any syncConfiguration so removing it seems kind of weird since it wasn't there in the first place. It would be nice if the initial config was `DEVICE_TO_CLOUD` then you remove it and then changes to `BETWEEN_DEVICE_AND_CLOUD` and then change it again to `DEVICE_TO_CLOUD` that way we know

1) It sets the correct value based on the original configuration (use non default value)
2) It responds correctly when removing a value by reverting back to its default value
3) When there is no existing value and you add one then it updates it to the value you want

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -171,4 +177,34 @@ void GIVEN_existing_sync_information_WHEN_config_updates_THEN_removed_sync_infor
...
                 new Pair<>(THING_NAME2, "Shadow-5")));
     }
+
+
+    @Test
+    @SuppressWarnings("PMD.CloseResource")
+    void GIVEN_shadow_manager_WHEN_individual_config_resets_THEN_respond_to_config_updates(ExtensionContext context)
+            throws Exception {
+        ignoreExceptionOfType(context, SkipSyncRequestException.class);
+        MqttClient mqttClient = mock(MqttClient.class);
```
Nelsonochoam, line 187: Note for the future: (maybe add a TODO) - We should use Fakes/Test Doubles for integration tests instead of mocking ( I know the test above is doing the same)

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
Nelsonochoam: If we are going to do it later, lets add a TODO

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: In the code, new node is added with the default value so we're able to remove. But yes, I agree with your comment, so updated the PR.

Nelsonochoam (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -353,7 +360,7 @@ void GIVEN_sync_config_and_no_cloud_WHEN_startup_THEN_cloud_version_updated_via_
                 .mockCloud(true)
                 .mockDao(true)
                 .build());
-
+        cdl.await(2,TimeUnit.SECONDS);
```
MikeDombo, line 363: this doesn't validate that we successfully counted down. Use assertTrue to verify that.

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -697,7 +704,7 @@ void GIVEN_cloud_update_request_WHEN_retryable_thrown_AND_new_cloud_update_reque
 
         // Fire the initial request
         handler.get().handleRequest(request1, "DoAll");
-
+        cdl.await(2, TimeUnit.SECONDS);
```
MikeDombo, line 707: same

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,178 @@ protected void install(InstallConfig installConfig) {
         }
 
         inboundRateLimiter.clear();
+        config.lookupTopics(CONFIGURATION_CONFIG_KEY).subscribe((what, newv) -> {
+            if (what.equals(WhatHappened.timestampUpdated)) {
+                return;
+            }
+            if (installConfig.configureSynchronizeConfig) {
+                configureSynchronization(newv);
```
MikeDombo, line 247: `newv` will be null on the first call. Can we avoid passing null? Just call these configure methods with no parameters and let them do the right thing by reading the current config state.

saranyailla: Since we're subscribing to the top level configuration key, running all of these configure methods even when their keys don't change is not needed. 

I'm handling null in the configure methods but I do get your comment on not passing null.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,178 @@ protected void install(InstallConfig installConfig) {
...
-                        inboundRateLimiter.setRate(maxLocalShadowUpdatesPerThingPerSecond);
-                    }
-                } catch (InvalidConfigurationException e) {
-                    serviceErrored(e);
-                }
-            });
+    private void configureSynchronization(Node newv) {
+        if (newv != null && !newv.childOf(CONFIGURATION_SYNCHRONIZATION_TOPIC)) {
+            return;
+        }
```
rbattle, line 277: You don't use `newv` in here for anything - why pass it in? This check could happen outside of `configureSynchronization` 

Passing it in makes it seem like this node is going to be used for getting the configuration

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -239,166 +239,178 @@ protected void install(InstallConfig installConfig) {
...
+                .dflt(DEFAULT_DOCUMENT_SIZE));
+        try {
+            Validator.validateMaxShadowSize(newMaxShadowSize);
+            Validator.setMaxShadowDocumentSize(newMaxShadowSize);
+            logger.atDebug()
+                    .setEventType("config")
+                    .kv("maxShadowSize", newMaxShadowSize)
+                    .log();
+        } catch (InvalidConfigurationException e) {
+            serviceErrored(new InvalidConfigurationException(e));
```
rbattle, line 358: does this need to create a `new InvalidConfigurationException`

Should these configureMethods just throw InvalidConfiguration and you can let the method that calls them move the service to `serviceErrored`?

saranyailla: In this PR, I'm only addressing the subscription on top level configuration key vs individual keys. Nothing inside of these methods is actually modified. I'd prefer to refactor the code in the later PRs if needed.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
saranyailla: That's how it was in the previous version of the PR where the check happened outside of the configure methods. Modified like this to address your comment about the config validation being a bit defensive so the configure methods check based on newv.

rbattle: got it - I reread the previous thread and can see why you made that change. I was thinking about a larger change where the synchandler and the rest reads their own config and then getting rid of that installConfig object

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
saranyailla: Updated.

rbattle (approved):
Thanks for making the changes. Based on our discussion, I'd like to see more of a refactor in a subsequent PR so that we can avoid converting so much of the config to data objects in the ShadowManager class.  The installConfig object also seems unnecessary

Nelsonochoam (approved):

PR #132, opened 2022-11-15
refactor: rate limits configuration updates

saranyailla:
**Issue #, if available:**

**Description of changes:**
Read and trigger configuration updates from the component configuration object whenever it is updated. This PR only moves rate limits configuration. 

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -269,7 +269,7 @@ public class ShadowManager extends PluginService {
 
     private void configureRateLimits(RateLimitsConfiguration rateLimitsConfig) {
         inboundRateLimiter.updateRateLimits(rateLimitsConfig.getMaxTotalLocalRequestRate(),
-                rateLimitsConfig.getMaxTotalLocalRequestRate());
+                rateLimitsConfig.getMaxLocalRequestRatePerThing());
         iotDataPlaneClientWrapper.updateRateLimits(rateLimitsConfig.getMaxOutboundUpdatesPerSecond());
     }
 
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -17,6 +17,7 @@ import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.lifecyclemanager.PluginService;
 import com.aws.greengrass.mqttclient.CallbackEventManager;
 import com.aws.greengrass.mqttclient.MqttClient;
+import com.aws.greengrass.shadowmanager.configuration.ComponentConfiguration;
 import com.aws.greengrass.shadowmanager.exception.InvalidConfigurationException;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowIPCHandler;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowRequestHandler;
@@ -65,10 +66,6 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_DOC_SIZE_LIMIT_B_TOPIC;
-import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_LOCAL_REQUESTS_RATE_PER_THING_TOPIC;
-import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_OUTBOUND_UPDATES_PS_TOPIC;
-import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_TOTAL_LOCAL_REQUESTS_RATE;
-import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_RATE_LIMITS_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_STRATEGY_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_SYNCHRONIZATION_TOPIC;
 import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_SYNC_DIRECTION_TOPIC;
@@ -91,6 +88,7 @@ public class ShadowManager extends PluginService {
     private final AuthorizationHandlerWrapper authorizationHandlerWrapper;
     private final InboundRateLimiter inboundRateLimiter;
     private final DeviceConfiguration deviceConfiguration;
+    private ComponentConfiguration componentConfiguration;
     @Getter
     private final DeleteThingShadowRequestHandler deleteThingShadowRequestHandler;
     @Getter
@@ -244,11 +242,9 @@ public class ShadowManager extends PluginService {
             if (what.equals(WhatHappened.timestampUpdated) || what.equals(WhatHappened.interiorAdded)) {
                 return;
             }
+            onConfigurationUpdate();
             if (installConfig.configureSynchronizeConfig) {
                 configureSynchronization(newv);
-            }
-            if (installConfig.configureRateLimitsConfig) {
-                configureRateLimits(newv);
             }
             if (installConfig.configureMaxDocSizeLimitConfig) {
                 configureMaxSizeDocLimitConfig(newv);
@@ -260,6 +256,14 @@ public class ShadowManager extends PluginService {
                 configureStrategy(newv);
             }
         });
+    }
+
+    private void onConfigurationUpdate() {
+        try {
+            componentConfiguration = ComponentConfiguration.from(componentConfiguration, getConfig());
+        } catch (InvalidConfigurationException e) {
+            serviceErrored(e);
+        }
     }
 
     Strategy replaceStrategyIfNecessary(Strategy currentStrategy, Strategy strategy) {
@@ -303,38 +307,6 @@ public class ShadowManager extends PluginService {
                 stopSyncingShadows(true);
                 startSyncingShadows(StartSyncInfo.builder().reInitializeSyncInfo(true).startSyncStrategy(true)
                         .updateCloudSubscriptions(true).build());
-            }
-        } catch (InvalidConfigurationException e) {
-            serviceErrored(e);
-        }
-    }
-
-    private void configureRateLimits(Node newv) {
-        if (newv != null && !newv.childOf(CONFIGURATION_RATE_LIMITS_TOPIC)) {
-            return;
-        }
-        Topics rateLimitTopics = config.lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_RATE_LIMITS_TOPIC);
-        Map<String, Object> rateLimitsPojo = rateLimitTopics.toPOJO();
-        try {
-            if (rateLimitsPojo.containsKey(CONFIGURATION_MAX_OUTBOUND_UPDATES_PS_TOPIC)) {
-                int maxOutboundSyncUpdatesPerSecond = Coerce.toInt(rateLimitsPojo
-                        .get(CONFIGURATION_MAX_OUTBOUND_UPDATES_PS_TOPIC));
-                Validator.validateOutboundSyncUpdatesPerSecond(maxOutboundSyncUpdatesPerSecond);
-                iotDataPlaneClientWrapper.setRate(maxOutboundSyncUpdatesPerSecond);
-            }
-
-            if (rateLimitsPojo.containsKey(CONFIGURATION_MAX_TOTAL_LOCAL_REQUESTS_RATE)) {
-                int maxTotalLocalRequestRate = Coerce.toInt(rateLimitsPojo
-                        .get(CONFIGURATION_MAX_TOTAL_LOCAL_REQUESTS_RATE));
-                Validator.validateTotalLocalRequestRate(maxTotalLocalRequestRate);
-                inboundRateLimiter.setTotalRate(maxTotalLocalRequestRate);
-            }
-
-            if (rateLimitsPojo.containsKey(CONFIGURATION_MAX_LOCAL_REQUESTS_RATE_PER_THING_TOPIC)) {
-                int maxLocalShadowUpdatesPerThingPerSecond = Coerce.toInt(rateLimitsPojo
-                        .get(CONFIGURATION_MAX_LOCAL_REQUESTS_RATE_PER_THING_TOPIC));
-                Validator.validateLocalShadowRequestsPerThingPerSecond(maxLocalShadowUpdatesPerThingPerSecond);
-                inboundRateLimiter.setRate(maxLocalShadowUpdatesPerThingPerSecond);
             }
         } catch (InvalidConfigurationException e) {
             serviceErrored(e);
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -18,6 +18,7 @@ import com.aws.greengrass.lifecyclemanager.PluginService;
 import com.aws.greengrass.mqttclient.CallbackEventManager;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.shadowmanager.configuration.ComponentConfiguration;
+import com.aws.greengrass.shadowmanager.configuration.RateLimitsConfiguration;
 import com.aws.greengrass.shadowmanager.exception.InvalidConfigurationException;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowIPCHandler;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowRequestHandler;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ipc/InboundRateLimiter.java
```diff
@@ -71,22 +103,18 @@ public void clear() {
...
-    /**
-     * Sets each inbound rate limiter per thing to specified rate.
-     *
-     * @param rate Max inbound requests per second per thing
-     */
-    public void setRate(int rate) {
-        ratePerThing.set(rate);
-        rateLimitersPerThing.forEach((k, v) -> v.setRate(ratePerThing.get()));
+    public Map<String, IpcRateLimiter> getRateLimitersPerThing() {
+        rateLimitersPerThing.forEach((k, v) -> v.setRate(getRatePerThing()));
```
MikeDombo, line 117: does `setRate` need to be called on config changes? If so, how are we doing that?

src/main/java/com/aws/greengrass/shadowmanager/ipc/InboundRateLimiter.java
```diff
@@ -71,22 +103,18 @@ public void clear() {
...
+     * @return totalInboundRateLimiter
      */
-    public void setTotalRate(int rate) {
-        totalRate.set(rate);
-        totalInboundRateLimiter.setRate(totalRate.get());
+    public IpcRateLimiter getTotalInboundRateLimiter() {
+        synchronized (totalInboundRateLimiter) {
+            totalInboundRateLimiter.setRate(getTotalRate());
+        }
+        return totalInboundRateLimiter;
```
rbattle, line 113: why does this need to be synchronized? `setRate` is already a synchronized method in the rate limiter

also this behavior could be very confusing - especially with no documentation. Calling a getter is going to potentially mutate the rate limiter. 

Is there a performance impact of looking at the config on every request? Why not subscribe to the config topic and only update on changes like in other config based objects?

src/main/java/com/aws/greengrass/shadowmanager/configuration/ComponentConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+package com.aws.greengrass.shadowmanager.configuration;
+
+import com.aws.greengrass.config.Topics;
+
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+
+public class ComponentConfiguration {
+    RateLimitsConfiguration rateLimitsConfiguration;
+
+    public ComponentConfiguration(RateLimitsConfiguration rateLimitsConfiguration) {
```
Nelsonochoam, line 15:
```suggestion
    private ComponentConfiguration(RateLimitsConfiguration rateLimitsConfiguration) {
```

Otherwise people can skip the creation logic on the factory method to properly create an instance of the configuration

src/main/java/com/aws/greengrass/shadowmanager/configuration/ComponentConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public static ComponentConfiguration from(ComponentConfiguration oldConfiguration, Topics updatedTopics) {
+        Topics serviceConfiguration = updatedTopics.lookupTopics(CONFIGURATION_CONFIG_KEY);
+        RateLimitsConfiguration rateLimitsConfiguration = RateLimitsConfiguration.from(serviceConfiguration);
+        ComponentConfiguration newConfiguration = new ComponentConfiguration(rateLimitsConfiguration);
+        newConfiguration.triggerUpdates(oldConfiguration);
+        return newConfiguration;
+    }
+
+    private void triggerUpdates(ComponentConfiguration oldConfiguration) {
```
Nelsonochoam, line 33: We could remove this method for now not adding any extra value. (see comment below)

src/main/java/com/aws/greengrass/shadowmanager/configuration/ComponentConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+        if (this.equals(oldConfiguration)) {
+            return;
+        }
+        if (hasRateLimitsChanged(oldConfiguration)) {
+            rateLimitsConfiguration.triggerUpdates();
+        }
+    }
+
+    private boolean hasRateLimitsChanged(ComponentConfiguration oldConfiguration) {
```
Nelsonochoam, line 42: Also, this could be removed

src/main/java/com/aws/greengrass/shadowmanager/configuration/ComponentConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+        newConfiguration.triggerUpdates(oldConfiguration);
+        return newConfiguration;
+    }
+
+    private void triggerUpdates(ComponentConfiguration oldConfiguration) {
+        if (this.equals(oldConfiguration)) {
+            return;
+        }
+        if (hasRateLimitsChanged(oldConfiguration)) {
+            rateLimitsConfiguration.triggerUpdates();
```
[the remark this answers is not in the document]
saranyailla: Shadow manager respond to some of its configuration updates based on the previous state. Here it is not needed, but I'd still want to keep a way to compare previous state and then update only when it is needed.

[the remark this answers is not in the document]
Nelsonochoam: You can still do that as part of the creation logic of this class.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Creates a new rate limits configuration object and triggers updates based on previous configuration.
+     *
+     * @param oldConfiguration previous configuration of the component
+     * @param serviceTopics    current configuration topics
+     * @return rate limits configuration objects
+     */
+    public static RateLimitsConfiguration from(ComponentConfiguration oldConfiguration, Topics serviceTopics) {
```
Nelsonochoam, line 89: Pass the `RateLimitsConfiguration` directly. This method does not need to depend on all the values from `ComponentConfiguration`

src/main/java/com/aws/greengrass/shadowmanager/configuration/ComponentConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Constructor for shadow manager component.
+     * @param oldConfiguration previous configuration of the component
+     * @param updatedTopics current configuration topics
+     * @return shadow manager component configuration object
+     */
+    public static ComponentConfiguration from(ComponentConfiguration oldConfiguration, Topics updatedTopics) {
+        Topics serviceTopics = updatedTopics.lookupTopics(CONFIGURATION_CONFIG_KEY);
+        RateLimitsConfiguration rateLimitsConfiguration = RateLimitsConfiguration.from(oldConfiguration, serviceTopics);
```
Nelsonochoam, line 27: pass the `RateLimitsConfiguration`.
```suggestion
        RateLimitsConfiguration rateLimitsConfiguration = RateLimitsConfiguration.from(getRateLimitsConfiguration(oldConfiguration), serviceTopics);
        
      ....
        
  private static RateLimitsConfiguration getRateLimitsConfiguration(ComponentConfiguration oldConfiguration) {
     if (oldConfiguration != null) {
         return oldConfiguration.getRateLimitsConfig();
    }

    return null;
}
```

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+        Validator.validateOutboundSyncUpdatesPerSecond(maxOutboundUpdatesPerSecond);
+        return maxOutboundUpdatesPerSecond;
+    }
+
+    private RateLimitsConfiguration(int maxLocalRequestRatePerThing,
+                                    int maxTotalLocalRequestRate,
+                                    int maxOutboundUpdatesPerSecond,
+                                    InboundRateLimiter inboundRateLimiter,
+                                    IotDataPlaneClientWrapper iotDataPlaneClientWrapper) {
+
```
Nelsonochoam, line 62:
```suggestion
```

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+        this.iotDataPlaneClientWrapper = iotDataPlaneClientWrapper;
+    }
+
+    private static RateLimitsConfiguration getRateLimitsConfigurationFromTopics(Topics topics) {
+        Topics rateLimitsTopics = topics.lookupTopics(CONFIGURATION_RATE_LIMITS_TOPIC);
+        int maxTotalLocalRequestRate = getMaxTotalLocalRequestRateFromTopics(rateLimitsTopics);
+        int maxLocalRequestRatePerThing = getMaxLocalRequestRatePerThingFromTopics(rateLimitsTopics);
+        int maxOutboundUpdatesPerSecond = getMaxOutboundUpdatesPerSecondFromTopics(rateLimitsTopics);
+        InboundRateLimiter inboundRateLimiter = topics.getContext().get(InboundRateLimiter.class);
+        IotDataPlaneClientWrapper iotDataPlaneClientWrapper = topics.getContext().get(IotDataPlaneClientWrapper.class);
```
Nelsonochoam, line 76: Seems weird that the configuration object depends on a `Client`. Seems like this dependencies should be inverted, in fact both `InboundRateLimiter` and the `InboundRateLimiter` depend on the configuration not the other way around. If we relied on events that would be an easier change but for now what we could do is:

1. Move the `InboundRateLimiter` and `IotDataPlaneClientWrapper` update logic to [here](IotDataPlaneClientWrapper)

2. Check if the configuration for the rate limited changed by

```
           componentConfiguration = ComponentConfiguration.from(componentConfiguration, getConfig());
           rateLimitConfig = componentConfiguration.getRateLimitConfiguration();

          if (rateLimitConfig.hasChanged(componentConfiguration.getRateLimitConfig()) {
               // Crate a method here to update those classes or just inline update them here
          }

```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -91,6 +88,7 @@ public class ShadowManager extends PluginService {
     private final AuthorizationHandlerWrapper authorizationHandlerWrapper;
     private final InboundRateLimiter inboundRateLimiter;
     private final DeviceConfiguration deviceConfiguration;
+    private ComponentConfiguration componentConfiguration;
```
MikeDombo, line 91: how is this being used? You are updating this field but never reading from it.

saranyailla: This is used in L263 `from` where the last updated value of `componentConfiguration` is used as the old state. For the first time, it is null.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+     * @return rate limits configuration objects
+     */
+    public static RateLimitsConfiguration from(RateLimitsConfiguration oldRateLimitsConfig, Topics serviceTopics) {
+        RateLimitsConfiguration rateLimitsConfiguration = getRateLimitsConfigurationFromTopics(serviceTopics);
+        rateLimitsConfiguration.triggerUpdates(oldRateLimitsConfig);
+        return rateLimitsConfiguration;
+    }
+
+    private void triggerUpdates(RateLimitsConfiguration oldRateLimitsConfiguration) {
+        if (hasMaxLocalRequestPerThingChanged(oldRateLimitsConfiguration)) {
```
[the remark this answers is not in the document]
saranyailla: Yes, here it can be updated without checking previous values. 

And for inbound rate limiter to read values directly from the config object, are you suggesting that I put `ComponentConfiguration` in context, so other classes can read the values from there?

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
saranyailla: Updated the PR by moving these back in main ShadowManager file.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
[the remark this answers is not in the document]
MikeDombo: Essentially yes. You should have a look at what @jbutler implemented in CDA regarding distributing configuration to where it is needed.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -262,6 +259,20 @@ protected void install(InstallConfig installConfig) {
...
+        try {
+            componentConfiguration = ComponentConfiguration.from(componentConfiguration, getConfig());
+            configureRateLimits(componentConfiguration.getRateLimitsConfiguration());
+        } catch (InvalidConfigurationException e) {
+            serviceErrored(e);
+        }
+    }
+
+    private void configureRateLimits(RateLimitsConfiguration rateLimitsConfiguration) {
+        inboundRateLimiter.updateRateLimits(rateLimitsConfiguration);
```
Nelsonochoam, line 272: Don't pass the entire config to the method just pass the values it needs (less coupling)

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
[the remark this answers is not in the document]
saranyailla: Yup. In CDA, we're using domain events, listeners and handlers based on config updates. Will do those iteratively as needed.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_OUTBOUND_UPDATES_PS_TOPIC;
+import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_MAX_TOTAL_LOCAL_REQUESTS_RATE;
+import static com.aws.greengrass.shadowmanager.model.Constants.CONFIGURATION_RATE_LIMITS_TOPIC;
+import static com.aws.greengrass.shadowmanager.model.Constants.DEFAULT_LOCAL_REQUESTS_RATE;
+import static com.aws.greengrass.shadowmanager.model.Constants.DEFAULT_MAX_OUTBOUND_SYNC_UPDATES_PS;
+import static com.aws.greengrass.shadowmanager.model.Constants.DEFAULT_TOTAL_LOCAL_REQUESTS_RATE;
+
+public final class RateLimitsConfiguration {
+    @Getter
+    private final int maxTotalLocalRequestRate;
```
MikeDombo, line 23: This information already exists in the config. This is what I wanted this change to remove. I don't want these fields to exist at all, rather have a getter which simply reads directly from the config.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+        int maxTotalLocalRequestRate = Coerce.toInt(rateLimitsTopics
+                .lookup(CONFIGURATION_MAX_TOTAL_LOCAL_REQUESTS_RATE)
+                .dflt(DEFAULT_TOTAL_LOCAL_REQUESTS_RATE));
+        Validator.validateTotalLocalRequestRate(maxTotalLocalRequestRate);
+        return maxTotalLocalRequestRate;
+    }
+
+    private static int getMaxLocalRequestRatePerThingFromTopics(Topics rateLimitsTopics) {
+        int maxLocalRequestRatePerThing = Coerce.toInt(rateLimitsTopics
+                .lookup(CONFIGURATION_MAX_LOCAL_REQUESTS_RATE_PER_THING_TOPIC)
```
MikeDombo, line 39: Prefer to use find or default rather than dflt. Dflt sets an initial value, but that value wont act like the default that you're expecting. Ie. It will not necessarily be used when the config value is missing.

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Creates a new rate limits configuration object and triggers updates based on previous configuration.
+     *
+     * @param serviceTopics    current configuration topics
+     * @return rate limits configuration objects
+     */
+    public static RateLimitsConfiguration from(Topics serviceTopics) {
+        return getRateLimitsConfigurationFromTopics(serviceTopics);
```
Nelsonochoam, line 78: You could just inline the implementation of this method here

Nelsonochoam (approved):

src/main/java/com/aws/greengrass/shadowmanager/configuration/RateLimitsConfiguration.java
saranyailla: Updated.

MikeDombo (approved):

Nelsonochoam (dismissed):

MikeDombo (approved):

PR #133, opened 2022-11-30
fix: set correct strategy when the config is updated

saranyailla:
**Issue #, if available:**

**Description of changes:**
When the strategy config is updated from periodic to real time (can also happen during reset), overall sync strategy is not set to real time. This fix properly updates the strategy after comparing with the current strategy. 

Added an integration test that fails without this fix. 

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -15,6 +15,8 @@ import com.aws.greengrass.shadowmanager.model.LogEvents;
 import com.aws.greengrass.shadowmanager.model.dao.SyncInformation;
 import com.aws.greengrass.shadowmanager.sync.SyncHandler;
 import com.aws.greengrass.shadowmanager.sync.model.Direction;
+import com.aws.greengrass.shadowmanager.sync.strategy.PeriodicSyncStrategy;
+import com.aws.greengrass.shadowmanager.sync.strategy.RealTimeSyncStrategy;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
 import com.aws.greengrass.util.Pair;
 import org.flywaydb.core.api.FlywayException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -102,6 +102,7 @@ public class ShadowManager extends PluginService {
     private final CloudDataClient cloudDataClient;
     private final MqttClient mqttClient;
     private final PubSubIntegrator pubSubIntegrator;
+    private final AtomicReference<Strategy> currentStrategy = new AtomicReference<>(DEFAULT_STRATEGY);
```
rbattle, line 105: this fixes the issue but lets add this to the list of things to refactor - we've got the state not copied in ShadowManager and SyncHandler

rbattle (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -210,4 +214,35 @@ void GIVEN_shadow_manager_WHEN_individual_config_resets_THEN_respond_to_config_u
...
+
+        kernel.getContext().put(MqttClient.class, mqttClient);
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile("periodic_sync.yaml")
+                .mqttConnected(false)
+                .syncClazz(PeriodicSyncStrategy.class)
+                .build());
+
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+        assertThat(syncHandler.getOverallSyncStrategy(), instanceOf(PeriodicSyncStrategy.class));
```
Nelsonochoam, line 235: I know we do this in another tests but this type of testing couples tests to implementation. The tests now knows that we use a class called `SyncHandler` and it also know that the `syncHandler` has a method called `getOverallSyncStrategy` and we are asserting a specific class gets returned making it hard to change.

Nelsonochoam (approved):

PR #135, opened 2022-12-05
refactor: wait for the crypto key provider service to load before con…

saranyailla:
…figuring the client

**Issue #, if available:**

**Description of changes:**
- Wait for the crypto key provider service to be available before configuring the client by retrying for 3 times. If the service is still not available, then throw an exception. 
- The data plane client is removed when the any of the device config changes. So that getter on data plane client re-configures it waiting for the crypto key provider service to load. If the client already exists, we return the same in the getter without reconfiguring. 
- With this an exception called  `IoTDataPlaneClientCreationException` (newly added with this PR) will be thrown when the client is not properly configured.
- Any cloud request that throws `IoTDataPlaneClientCreationException` will be retried as the client is not configured.

If the data plane client is configured before the crypto key provider service (such as pkcs provider component) is loaded, then data plane client will not be mTLS [configured.](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a8ff7f53eac7a301103969fcbbb2d567e3ccdf8/src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java#L105) This will make the requests using SDK throw SDKClientException due to SSLHandShakeException. These requests are always skipped and never retried.

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
@@ -8,6 +8,7 @@ package com.aws.greengrass.integrationtests;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
 import com.aws.greengrass.shadowmanager.ShadowManagerDatabase;
 import com.aws.greengrass.shadowmanager.model.ShadowDocument;
@@ -19,9 +20,11 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import vendored.com.google.common.util.concurrent.RateLimiter;
 
+import javax.net.ssl.KeyManager;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileVisitResult;
@@ -62,14 +65,19 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.lenient;
 
 @ExtendWith({MockitoExtension.class, GGExtension.class})
 @SuppressWarnings("PMD.CloseResource")
 class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
     ShadowManagerDatabase db;
+    @Mock
+    SecurityService securityService;
+
 
     @BeforeEach
     void initializeShadowManagerDatabase() {
+        System.setProperty("aws.greengrass.scanSelfClasspath", "true");
         db = new ShadowManagerDatabase(rootDir);
         db.install();
         db.open();
@@ -82,7 +90,9 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
 
     @Test
     void GIVEN_data_WHEN_restart_THEN_data_still_exists() throws Exception {
+        lenient().when(securityService.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
         kernel = new Kernel();
+        kernel.getContext().put(SecurityService.class, securityService);
         try {
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
             ShadowManagerDatabase shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
@@ -110,6 +110,7 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
             kernel.shutdown();
 
             kernel = new Kernel();
+            kernel.getContext().put(SecurityService.class, securityService);
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
 
             shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
@@ -24,6 +24,7 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import vendored.com.google.common.util.concurrent.RateLimiter;
 
+import javax.net.ssl.KeyManager;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileVisitResult;
@@ -64,6 +65,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.lenient;
 
 @ExtendWith({MockitoExtension.class, GGExtension.class})
 @SuppressWarnings("PMD.CloseResource")
@@ -88,7 +90,9 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
 
     @Test
     void GIVEN_data_WHEN_restart_THEN_data_still_exists() throws Exception {
+        lenient().when(securityService.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
         kernel = new Kernel();
+        kernel.getContext().put(SecurityService.class, securityService);
         try {
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
             ShadowManagerDatabase shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
@@ -24,7 +24,6 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import vendored.com.google.common.util.concurrent.RateLimiter;
 
-import javax.net.ssl.KeyManager;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileVisitResult;
@@ -65,7 +64,6 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
-import static org.mockito.Mockito.lenient;
 
 @ExtendWith({MockitoExtension.class, GGExtension.class})
 @SuppressWarnings("PMD.CloseResource")
@@ -90,9 +88,7 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
 
     @Test
     void GIVEN_data_WHEN_restart_THEN_data_still_exists() throws Exception {
-        lenient().when(securityService.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
         kernel = new Kernel();
-        kernel.getContext().put(SecurityService.class, securityService);
         try {
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
             ShadowManagerDatabase shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
@@ -8,7 +8,6 @@ package com.aws.greengrass.integrationtests;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.impl.LogManager;
-import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
 import com.aws.greengrass.shadowmanager.ShadowManagerDatabase;
 import com.aws.greengrass.shadowmanager.model.ShadowDocument;
@@ -20,11 +19,9 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import vendored.com.google.common.util.concurrent.RateLimiter;
 
-import javax.net.ssl.KeyManager;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileVisitResult;
@@ -65,15 +62,11 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
-import static org.mockito.Mockito.lenient;
 
 @ExtendWith({MockitoExtension.class, GGExtension.class})
 @SuppressWarnings("PMD.CloseResource")
 class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
     ShadowManagerDatabase db;
-    @Mock
-    SecurityService securityService;
-
 
     @BeforeEach
     void initializeShadowManagerDatabase() {
@@ -90,9 +83,7 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
 
     @Test
     void GIVEN_data_WHEN_restart_THEN_data_still_exists() throws Exception {
-        lenient().when(securityService.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
         kernel = new Kernel();
-        kernel.getContext().put(SecurityService.class, securityService);
         try {
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
             ShadowManagerDatabase shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
@@ -110,7 +101,6 @@ class ShadowManagerDatabaseTest extends NucleusLaunchUtils {
             kernel.shutdown();
 
             kernel = new Kernel();
-            kernel.getContext().put(SecurityService.class, securityService);
             startNucleusWithConfig("config.yaml", State.RUNNING, false);
 
             shadowManagerDatabase = kernel.getContext().get(ShadowManagerDatabase.class);
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -9,7 +9,6 @@ import com.aws.greengrass.authorization.exceptions.AuthorizationException;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.mqttclient.MqttClient;
-import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
 import com.aws.greengrass.shadowmanager.exception.SkipSyncRequestException;
 import com.aws.greengrass.shadowmanager.model.LogEvents;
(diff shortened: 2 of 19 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -87,12 +72,20 @@ private String getIotCoreDataPlaneEndpoint(String iotDataEndpoint) {
...
-    }
-
-    private void configureClient(DeviceConfiguration deviceConfiguration) {
+        // To ensure that the http client is configured with mutual auth, wait for the crypto key provider service
+        // to load. If the service is not loaded even after retrying, we continue with the client creation as shadow
+        // manager can still work locally.
+        try {
+            waitForCryptoKeyServiceProvider();
+        } catch (Exception e) {
+            logger.atError().log("Failed to get the key managers from crypto key provider service. Data plane"
```
MikeDombo, line 86: why are we returning an invalid client, ever?

Shadow manager will need to backoff and retry, waiting until the client is properly configured.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -68,10 +68,11 @@ class ShadowManagerTest extends NucleusLaunchUtils {
     public static final String THING_NAME2 = "testThingName2";
 
     @BeforeEach
-    void setup() {
+    void setup() throws Exception {
         // Set this property for kernel to scan its own classpath to find plugins
         System.setProperty("aws.greengrass.scanSelfClasspath", "true");
         kernel = new Kernel();
+        lenient().doNothing().when(iotDataPlaneClientFactory).waitForCryptoKeyServiceProvider();
```
jbutler, line 75: This is not something that should be mocked

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
jbutler: Agree that this class should be retrying anything or blocking. We don't know what thread context we're running in. Blocking a thread is not appropriate.

From an interface perspective, this class should only return a client if it is actually usable. We'll need to think carefully about how to handle this at the upper layers

src/test/java/com/aws/greengrass/shadowmanager/sync/model/FullShadowSyncRequestTest.java
```diff
@@ -6,10 +6,7 @@
 package com.aws.greengrass.shadowmanager.sync.model;
 
 import com.aws.greengrass.shadowmanager.ShadowManagerDAO;
-import com.aws.greengrass.shadowmanager.exception.InvalidRequestParametersException;
-import com.aws.greengrass.shadowmanager.exception.RetryableException;
-import com.aws.greengrass.shadowmanager.exception.ShadowManagerDataException;
-import com.aws.greengrass.shadowmanager.exception.SkipSyncRequestException;
+import com.aws.greengrass.shadowmanager.exception.*;
```
MikeDombo, line 9: no * imports

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -125,4 +111,38 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+    /**
+     * Checks if the crypto key service provider is available by getting key managers.
+     *
+     * @throws Exception exception during retry
+     */
+    @SuppressWarnings({"PMD.SignatureDeclareThrowsException"})
+    public void waitForCryptoKeyServiceProvider() throws Exception {
+        logger.atDebug().log("Checking if the crypto key service provider is available");
+        RetryUtils.RetryConfig retryConfig =
+                RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(2)).maxAttempt(3)
```
MikeDombo, line 141: do we really want to retry here? Or should we let it retry at the higher level?

I'm alright with this, just asking.

saranyailla: If we want to retry only for `TLSAuthException` type, this would be cleaner.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientWrapper.java
```diff
@@ -83,8 +88,10 @@ public UpdateThingShadowResponse updateThingShadow(String thingName, String shad
      *
      * @param thingName  The thing name associated with the sync shadow update
      * @param shadowName The shadow name associated with the sync shadow update
+     * @throws IoTDataPlaneClientCreationException when the iot data plane client is not created
      */
-    public GetThingShadowResponse getThingShadow(String thingName, String shadowName) {
+    public GetThingShadowResponse getThingShadow(String thingName, String shadowName)
+            throws IoTDataPlaneClientCreationException {
         rateLimiter.acquire();
```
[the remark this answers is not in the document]
saranyailla: Nice catch. But, can I change it in the next PR as a follow up as I am not sure if that'll update the tests as well?

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
MikeDombo: You're throwing TLSAuthException up to the caller too, so they can also retry on that specific error if they want.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientWrapper.java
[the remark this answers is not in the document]
MikeDombo: ok

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: Update the PR to throw a new exception called - `IoTDataPlaneClientCreation` when the service is not ready. All the cloud requests failing with this exception type are now retried.

MikeDombo (approved):

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -125,4 +106,41 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException"})
+    public IotDataPlaneClient getIotDataPlaneClient() throws IoTDataPlaneClientCreationException {
+        try {
+            // To ensure that the http client is configured with mTLS, wait for the crypto key provider service (pkcs11)
+            // to load. If the service is not loaded even after retrying, we throw an exception.
+            waitForCryptoKeyServiceProvider();
+        } catch (Exception e) {
+            throw new IoTDataPlaneClientCreationException(e);
+        }
+        configureClient();
```
jbutler, line 125: I'm curious how expensive this is. I _think_ we've already had some complaints about ShadowManager not being able to keep up (am I making stuff up?) and I'd be nervous about this worsening that.

src/test/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactoryTest.java
```diff
@@ -39,49 +40,42 @@ class IotDataPlaneClientFactoryTest {
...
+        lenient().when(mockDeviceConfiguration.getAWSRegion()).thenReturn(regionTopic);
     }
 
     @Test
-    void GIVEN_bad_config_WHEN_configure_client_THEN_does_not_reconfigure_iot_data_client() {
+    void GIVEN_device_configuration_WHEN_crypto_service_available_THEN_configure_iot_data_client() throws TLSAuthException, IoTDataPlaneClientCreationException {
         clientFactory = new IotDataPlaneClientFactory(mockDeviceConfiguration);
         assertThat(clientFactory.getIotDataPlaneClient(), is(notNullValue()));
-
+        lenient().when(mockDeviceConfiguration.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
```
jbutler, line 58: nit: is this called or not? Shouldn't need to use `lenient` here

jbutler (commented):
I don't have any problems with the changes here, but I'm not completely satisfied with the test coverage.

Mostly we're just updating unit tests with the new exception, but nowhere is the behavior actually tested.

What's the plan for that?

src/test/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactoryTest.java
saranyailla: This should be updated. And integ tests for these changes will be added in the next PR (unit test is in this PR).

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: I can revert it to the [previous state](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/main/src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java#L66-L72) where client is configured only in the constructor and when the `deviceConfiguration` changes. Here we can wait for the crypto key provider service to load.

jbutler: Re-using is always preferable as long as we're not opening up other holes. If we re-init every time then I'd want to do some perf testing to ensure it's not negatively impacting us.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -55,23 +59,23 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.lenient;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
```
MikeDombo, line 62: no * imports. Fix your IDE.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -63,12 +73,13 @@ public class IotDataPlaneClientFactory {
      */
     @Inject
     public IotDataPlaneClientFactory(DeviceConfiguration deviceConfiguration) {
-        configureClient(deviceConfiguration);
+        this.deviceConfiguration = deviceConfiguration;
+        configureClient(true);
```
MikeDombo, line 77: this may cause blocking in the constructor! We must not block

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -63,12 +73,13 @@ public class IotDataPlaneClientFactory {
...
         deviceConfiguration.onAnyChange((what, node) -> {
             if (validString(node, DEVICE_PARAM_AWS_REGION) || validPath(node, DEVICE_PARAM_ROOT_CA_PATH) || validPath(
                     node, DEVICE_PARAM_CERTIFICATE_FILE_PATH) || validPath(node, DEVICE_PARAM_PRIVATE_KEY_PATH)
                     || validString(node, DEVICE_PARAM_IOT_DATA_ENDPOINT)) {
-                configureClient(deviceConfiguration);
+                configureClient(true);
```
MikeDombo, line 82: this will block the main context thread due to retries. We must not block

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -125,4 +149,37 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+     * Getter for IoT data plane client. This configures the client everytime the getter is used.
+     *
+     * @return iotDataPlaneClient
+     * @throws IoTDataPlaneClientCreationException exception during client configuration
+     */
+    public IotDataPlaneClient getIotDataPlaneClient() throws IoTDataPlaneClientCreationException {
+        configureClient(clientCreationException.get().isPresent());
+        Optional<Exception> exception = clientCreationException.get();
+        if (exception.isPresent()) {
+            throw new IoTDataPlaneClientCreationException(exception.get());
```
MikeDombo, line 163: why are you just throwing the exception?

The key provider may be available by now. This method should try to get a configured client if we don't have one available.

saranyailla: Yeah, this is even after trying to configure the client again in L160.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: Totally removed client configuration in constructor and subscribe as suggested. Setting the client to null in any of these cases. May have to make it Atomic and optional.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
MikeDombo: OK. Not sure holding the exception is needed. Just using `iotDataPlaneClient == null` or not would seem to accomplish the same in a less complicated way.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
MikeDombo: Atomic+Optional is totally unnecessary

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -125,4 +149,37 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+     * @throws Exception exception during retry
+     */
+    @SuppressWarnings({"PMD.SignatureDeclareThrowsException"})
+    public void waitForCryptoKeyServiceProvider() throws Exception {
+        logger.atDebug().log("Checking if the crypto key service provider is available");
+        RetryUtils.RetryConfig retryConfig =
+                RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(2)).maxAttempt(3)
+                        .retryableExceptions(Collections.singletonList(TLSAuthException.class)).build();
+        // Ensures that the crypto key provider service is available as the key managers are obtained from it.
+        RetryUtils.runWithRetry(retryConfig,
```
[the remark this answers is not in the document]
saranyailla: But`runWithRetry` can also throw `InterruptedException` sometimes, right?

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: Removed.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: Removed exception.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: removed calls from constructor and subscribe as they block the main context thread.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/UnhappyUpdateIPCTest.java
```diff
@@ -7,7 +7,7 @@
 
 import com.aws.greengrass.integrationtests.NucleusLaunchUtils;
 import com.aws.greengrass.lifecyclemanager.Kernel;
-import com.aws.greengrass.shadowmanager.exception.InvalidRequestParametersException;
+ import com.aws.greengrass.shadowmanager.exception.InvalidRequestParametersException;
```
MikeDombo, line 10: undo

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -55,20 +60,22 @@ public class IotDataPlaneClientFactory {
...
+    @SuppressWarnings({"PMD.NullAssignment"})
     public IotDataPlaneClientFactory(DeviceConfiguration deviceConfiguration) {
-        configureClient(deviceConfiguration);
+        this.deviceConfiguration = deviceConfiguration;
         deviceConfiguration.onAnyChange((what, node) -> {
             if (validString(node, DEVICE_PARAM_AWS_REGION) || validPath(node, DEVICE_PARAM_ROOT_CA_PATH) || validPath(
                     node, DEVICE_PARAM_CERTIFICATE_FILE_PATH) || validPath(node, DEVICE_PARAM_PRIVATE_KEY_PATH)
                     || validString(node, DEVICE_PARAM_IOT_DATA_ENDPOINT)) {
-                configureClient(deviceConfiguration);
+                iotDataPlaneClient = null;
```
MikeDombo, line 78: now this isn't threadsafe, so you need to use an atomic or synchronize here.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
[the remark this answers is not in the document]
MikeDombo: If it is interrupted, we should handle that differently. Interrupt has particular semantics about it.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -125,4 +143,33 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
     }
+
+    /**
+     * Getter for IoT data plane client. This configures the client everytime the getter is used.
+     *
+     * @return iotDataPlaneClient
+     * @throws IoTDataPlaneClientCreationException exception during client configuration
+     */
+    public IotDataPlaneClient getIotDataPlaneClient() throws IoTDataPlaneClientCreationException {
+        configureClient();
```
MikeDombo, line 154: this isn't safe.

Configure sets the client in a field which you then return from. A different thread could run the callback which sets the client to null, now you just returned null to your caller.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
saranyailla: updated the PR with atomic client

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
MikeDombo: this still isn't safe. You're using atomic, but without any locking this is exactly equivalent in safety to what you had before.

Have configureClient return you a client and return it here to actually be safe.

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -120,9 +139,39 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
         }
-        if (this.iotDataPlaneClient != null) {
-            this.iotDataPlaneClient.close();
+        if (this.iotDataPlaneClient.get() != null) {
+            this.iotDataPlaneClient.get().close();
         }
-        this.iotDataPlaneClient = iotDataPlaneClientBuilder.build();
+        this.iotDataPlaneClient.set(iotDataPlaneClientBuilder.build());
+        return iotDataPlaneClient.get();
+
```
MikeDombo, line 147: this is still a race

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -92,7 +100,18 @@ private boolean validPath(Node node, String key) {
         return validString(node, key) && Files.exists(Paths.get(key));
     }
 
-    private void configureClient(DeviceConfiguration deviceConfiguration) {
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException"})
+    private IotDataPlaneClient configureAndGetClient() throws IoTDataPlaneClientCreationException {
+        if (iotDataPlaneClient.get() != null) {
+            return iotDataPlaneClient.get();
+        }
```
MikeDombo, line 107: this is a race.

`client = iotDataPlaneClient.get()`
`if (client != null) return client`

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -120,9 +139,40 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
         if (!Utils.isEmpty(iotDataEndpoint)) {
             iotDataPlaneClientBuilder.endpointOverride(URI.create(getIotCoreDataPlaneEndpoint(iotDataEndpoint)));
         }
-        if (this.iotDataPlaneClient != null) {
-            this.iotDataPlaneClient.close();
+        if (this.iotDataPlaneClient.get() != null) {
+            this.iotDataPlaneClient.get().close();
```
MikeDombo, line 143: this is a race

MikeDombo (approved):

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientFactory.java
```diff
@@ -120,9 +140,50 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+    }
+
+    /**
+     * Setter for IoT data plane client.
+     *
+     * @param newClient client to set as iot data plane client
+     */
+    @SuppressWarnings("PMD.CloseResource")
+    private void setIotDataPlaneClient(IotDataPlaneClient newClient) {
+        IotDataPlaneClient client = iotDataPlaneClient.get();
```
MikeDombo, line 165: just use getAndSet?

MikeDombo (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerDatabaseTest.java
```diff
@@ -82,7 +90,9 @@ void close() throws IOException {
 
     @Test
     void GIVEN_data_WHEN_restart_THEN_data_still_exists() throws Exception {
+        lenient().when(securityService.getDeviceIdentityKeyManagers()).thenReturn(new KeyManager[0]);
         kernel = new Kernel();
+        kernel.getContext().put(SecurityService.class, securityService);
```
jbutler, line 95: This test isn't hitting the cloud. Why do we need to mock this?

jbutler (approved):
This PR is already getting fairly "noisy" (lots of updated tests), so I don't want to make it any bigger.
But I would like to see a test that simulates the security service failing to start as a separate PR

jbutler (approved):

MikeDombo (approved):

PR #137, opened 2022-12-05
test: remove assertNotEmptySyncQueue check

saranyailla:
**Issue #, if available:**

**Description of changes:**
Fix two tests with possible race conditions. 
- In the tests GIVEN_cloud_sync_not_enabled_WHEN_local_update_THEN_does_not_sync_shadow_to_cloud, we check  if the sync queue is not empty right before checking if it's empty. It is not always guaranteed that both of them will hold true due to timing of the tests. So, remove checking if the queue is not empty completely instead check if the certain number of requests (all) are processed from that queue.
- In the GIVEN_synced_shadow_WHEN_local_update_THEN_cloud_updates, before making an IPC request, this change will make sure that all the requests before that are processed. 

**Checklist:**
 

rbattle (approved):
Please update the pull request description

Nelsonochoam (approved):

dgggIoT (approved):

PR #139, opened 2022-12-13
test: check that requests are retried when the crypto key service is still loading

saranyailla:
…add tests

**Issue #, if available:**
Addressing comments from #135 
- Add integ test without mocking cloud to verify that the requests fail when the client does not exist and then retried along with the creation of a new client. 

**Description of changes:**

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -293,7 +293,6 @@ class ShadowManagerTest extends NucleusLaunchUtils {
         BaseSyncStrategy syncStrategy = kernel.getContext().get(RealTimeSyncStrategy.class);
         assertThat("syncing has started", syncStrategy::isSyncing, eventuallyEval(is(true)));
         verify(wrapper, timeout(5000).atLeast(1)).getThingShadow("Thing1", "");
-        verify(securityService, timeout(5000).atLeast(3)).getDeviceIdentityKeyManagers();
 
         // This test ensures that the request is retried again as the we're not mocking cloud and data plane client
         // creation fails. If the data plane client creation succeeds, the request is processed once again.
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -271,8 +271,7 @@ class ShadowManagerTest extends NucleusLaunchUtils {
         ignoreExceptionOfType(context, RetryableException.class);
         ignoreExceptionOfType(context, ResourceNotFoundException.class);
 
-        IotDataPlaneClientFactory factory = spy(kernel.getContext().get(IotDataPlaneClientFactory.class));
-        kernel.getContext().put(IotDataPlaneClientFactory.class, factory);
+        IotDataPlaneClientFactory factory = kernel.getContext().get(IotDataPlaneClientFactory.class);
         IotDataPlaneClientWrapper wrapper = spy(new FakeIotDataPlaneClientWrapper(factory));
         kernel.getContext().put(IotDataPlaneClientWrapper.class, wrapper);
 
@@ -281,8 +280,8 @@ class ShadowManagerTest extends NucleusLaunchUtils {
                 .mqttConnected(true)
                 .mockCloud(false)
                 .build());
-        BaseSyncStrategy syncStragy = kernel.getContext().get(RealTimeSyncStrategy.class);
-        assertThat("syncing has started", s::isSyncing, eventuallyEval(is(true)));
+        BaseSyncStrategy syncStrategy = kernel.getContext().get(RealTimeSyncStrategy.class);
+        assertThat("syncing has started", syncStrategy::isSyncing, eventuallyEval(is(true)));
         verify(wrapper, timeout(5000).atLeast(1)).getThingShadow("Thing1", "");
 
         CountDownLatch cdl = new CountDownLatch(1);
@@ -294,6 +293,7 @@ class ShadowManagerTest extends NucleusLaunchUtils {
         kernel.getContext().put(SecurityService.class, ss);
         assertThat("request is retried with a new client",cdl.await(10, TimeUnit.SECONDS), is(true));
         verify(wrapper, timeout(5000).atLeast(2)).getThingShadow("Thing1", "");
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
     }
 
     static class FakeIotDataPlaneClientWrapper extends  IotDataPlaneClientWrapper{
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -281,7 +281,7 @@ class ShadowManagerTest extends NucleusLaunchUtils {
                 .mqttConnected(true)
                 .mockCloud(false)
                 .build());
-        BaseSyncStrategy s = kernel.getContext().get(RealTimeSyncStrategy.class);
+        BaseSyncStrategy syncStragy = kernel.getContext().get(RealTimeSyncStrategy.class);
         assertThat("syncing has started", s::isSyncing, eventuallyEval(is(true)));
         verify(wrapper, timeout(5000).atLeast(1)).getThingShadow("Thing1", "");
 
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -10,15 +10,18 @@ import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl;
+import com.aws.greengrass.shadowmanager.exception.IoTDataPlaneClientCreationException;
 import com.aws.greengrass.shadowmanager.exception.SkipSyncRequestException;
 import com.aws.greengrass.shadowmanager.model.LogEvents;
 import com.aws.greengrass.shadowmanager.model.dao.SyncInformation;
+import com.aws.greengrass.shadowmanager.sync.IotDataPlaneClientFactory;
 import com.aws.greengrass.shadowmanager.sync.SyncHandler;
 import com.aws.greengrass.shadowmanager.sync.model.Direction;
 import com.aws.greengrass.shadowmanager.sync.strategy.PeriodicSyncStrategy;
 import com.aws.greengrass.shadowmanager.sync.strategy.RealTimeSyncStrategy;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
 import com.aws.greengrass.util.Pair;
+import com.aws.greengrass.util.exceptions.TLSAuthException;
 import org.flywaydb.core.api.FlywayException;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -53,6 +56,7 @@ import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.atLeastOnce;
@@ -246,5 +250,17 @@ class ShadowManagerTest extends NucleusLaunchUtils {
         shadowManager.getConfig().lookupTopics(CONFIGURATION_CONFIG_KEY, CONFIGURATION_STRATEGY_TOPIC).replaceAndWait(periodicStrategy);
         kernel.getContext().waitForPublishQueueToClear();
         assertThat(syncHandler.getOverallSyncStrategy(), instanceOf(PeriodicSyncStrategy.class));
+    }
+
+    @Test
+    void GIVEN_cryptoKeyProviderService_is_not_loaded_WHEN_get_client_THEN_throw_Exception(ExtensionContext context) throws InterruptedException {
+        ignoreExceptionOfType(context, TLSAuthException.class);
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile(DEFAULT_CONFIG)
+                .mqttConnected(false)
+                .mockCloud(false)
+                .build());
+        IotDataPlaneClientFactory clientFactory = kernel.getContext().get(IotDataPlaneClientFactory.class);
+        assertThrows(IoTDataPlaneClientCreationException.class, ()-> clientFactory.getIotDataPlaneClient());
     }
 }
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
@@ -36,7 +36,6 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import software.amazon.awssdk.services.iotdataplane.model.GetThingShadowResponse;
 
 import javax.net.ssl.KeyManager;
-
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -72,6 +71,7 @@ import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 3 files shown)


MikeDombo (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +251,16 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
         kernel.getContext().waitForPublishQueueToClear();
         assertThat(syncHandler.getOverallSyncStrategy(), instanceOf(PeriodicSyncStrategy.class));
     }
+
+    @Test
```
jbutler, line 255: This is a unit test.

For integ, we should be ensuring that ShadowManager retries and succeeds after the security service loads

src/main/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientWrapper.java
```diff
@@ -57,10 +58,12 @@ public void setRate(int rate) {
      * @param shadowName The shadow name associated with the sync shadow update
      * @throws IoTDataPlaneClientCreationException when the iot data plane client is not created
      */
+    @SuppressWarnings("PMD.CloseResource")
     public DeleteThingShadowResponse deleteThingShadow(String thingName, String shadowName)
             throws IoTDataPlaneClientCreationException {
+        IotDataPlaneClient client = iotDataPlaneClientFactory.getIotDataPlaneClient();
         rateLimiter.acquire();
-        return iotDataPlaneClientFactory.getIotDataPlaneClient().deleteThingShadow(DeleteThingShadowRequest.builder()
+        return client.deleteThingShadow(DeleteThingShadowRequest.builder()
```
jbutler, line 66: Is this even worth doing? If we can't get a data client, then the rate limiter doesn't matter

saranyailla: Yes, with this, exception is thrown without acquiring rate limiter.

jbutler: It's just more code without any benefit though? I think this can be backed out

jbutler (approved):
Just update the commit message when you merge. This isn't fixing anything. Just adding tests

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+    void GIVEN_cryptoKeyProviderService_WHEN_get_client_THEN_wait_for_cryptoservice(ExtensionContext context) throws InterruptedException, TLSAuthException, IoTDataPlaneClientCreationException {
```
Nelsonochoam, line 269: nit: are we using snake or camel case for test method names?

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        when(ss.getDeviceIdentityKeyManagers()).thenAnswer((invocation)->{
+            cdl.countDown();
+            return new KeyManager[0];
+        });
+        kernel.getContext().put(SecurityService.class, ss);
+        assertThat("request is retried with a new client",cdl.await(10, TimeUnit.SECONDS), is(true));
+        verify(wrapper, timeout(5000).atLeast(2)).getThingShadow("Thing1", "");
+    }
+
+    static class FakeIotDataPlaneClientWrapper extends  IotDataPlaneClientWrapper{
```
Nelsonochoam, line 299: Let's move this into its own file, so it is more easily usable by other test cases in the future

```suggestion
    static class FakeIotDataPlaneClientWrapper extends IotDataPlaneClientWrapper {
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+                .configFile("sync.yaml")
+                .mqttConnected(true)
+                .mockCloud(false)
+                .build());
+        BaseSyncStrategy s = kernel.getContext().get(RealTimeSyncStrategy.class);
+        assertThat("syncing has started", s::isSyncing, eventuallyEval(is(true)));
+        verify(wrapper, timeout(5000).atLeast(1)).getThingShadow("Thing1", "");
+
+        CountDownLatch cdl=new CountDownLatch(1);
+        SecurityService ss= mock(SecurityService.class);
```
Nelsonochoam, line 289:
```suggestion
        SecurityService securityService = mock(SecurityService.class);
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
         assertThat(syncHandler.getOverallSyncStrategy(), instanceOf(PeriodicSyncStrategy.class));
     }
+
+    @Test
+    void GIVEN_cryptoKeyProviderService_WHEN_get_client_THEN_wait_for_cryptoservice(ExtensionContext context) throws InterruptedException, TLSAuthException, IoTDataPlaneClientCreationException {
+        ignoreExceptionOfType(context, TLSAuthException.class);
+        ignoreExceptionOfType(context, RetryableException.class);
+        ignoreExceptionOfType(context, ResourceNotFoundException.class);
+
+        IotDataPlaneClientFactory factory = spy(kernel.getContext().get(IotDataPlaneClientFactory.class));
```
Nelsonochoam, line 274: Why are we spying on this factory? - not seeing any assertions or mocks on the spy

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+    static class FakeIotDataPlaneClientWrapper extends  IotDataPlaneClientWrapper{
+        IotDataPlaneClientFactory factory;
+
+        public FakeIotDataPlaneClientWrapper(IotDataPlaneClientFactory iotDataPlaneClientFactory) {
+            super(iotDataPlaneClientFactory);
+            factory = iotDataPlaneClientFactory;
+        }
+
+        @Override
+        public GetThingShadowResponse getThingShadow(String thingName, String shadowName) throws IoTDataPlaneClientCreationException {
```
Nelsonochoam, line 308: We should go one level deeper instead of creating a fake at this level. Doing this will hide away potential failures on the Wrapper coed. We create and `IotDataPlaneClient` fake that throws the error you expect that way we are only faking the boundaries

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        kernel.getContext().put(IotDataPlaneClientFactory.class, factory);
+        IotDataPlaneClientWrapper wrapper = spy(new FakeIotDataPlaneClientWrapper(factory));
+        kernel.getContext().put(IotDataPlaneClientWrapper.class, wrapper);
+
+        startNucleusWithConfig(NucleusLaunchUtilsConfig.builder()
+                .configFile("sync.yaml")
+                .mqttConnected(true)
+                .mockCloud(false)
+                .build());
+        BaseSyncStrategy s = kernel.getContext().get(RealTimeSyncStrategy.class);
```
Nelsonochoam, line 284: Lets use more descriptive names

```suggestion
        BaseSyncStrategy syncStragy = kernel.getContext().get(RealTimeSyncStrategy.class);
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: Agreed!

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        assertThat("syncing has started", s::isSyncing, eventuallyEval(is(true)));
+        verify(wrapper, timeout(5000).atLeast(1)).getThingShadow("Thing1", "");
+
+        CountDownLatch cdl=new CountDownLatch(1);
```
Nelsonochoam, line 288:
```suggestion
        CountDownLatch cdl = new CountDownLatch(1);
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,51 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        SecurityService ss= mock(SecurityService.class);
+        when(ss.getDeviceIdentityKeyManagers()).thenAnswer((invocation)->{
+            cdl.countDown();
+            return new KeyManager[0];
+        });
+        kernel.getContext().put(SecurityService.class, ss);
```
Nelsonochoam, line 294: Seems like we are just testing the positive case here meaning the security service loads and we continue, but should we test the negated case? - Meaning, if the security service fails to load we should not see an extra call.

saranyailla: We're testing the case where requests should be retried until the security service is loaded. Once it is loaded the request is processed, and here, it throws `ResourceNotFoundException`.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: However, in this case, it gets tricky as creation of a new `IotDataPlaneClient` is part of the code in test itself and moreover that is an interface. It is not injected as such.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: Removed.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: I am now asserting that the sync queue is empty at the end of the tests since the service is loaded and the request is processed.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
Nelsonochoam: Yeah I see. We could still create a fake on the factory and only override the getClientMethod

Nelsonochoam (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +263,58 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+
+        // This test ensures that the request is retried again as the we're not mocking cloud and data plane client
+        // creation fails. If the data plane client creation succeeds, the request is processed once again.
+        CountDownLatch cdl = new CountDownLatch(1);
+        SecurityService securityService = mock(SecurityService.class);
+        when(securityService.getDeviceIdentityKeyManagers()).thenAnswer((invocation) -> {
+            cdl.countDown();
+            return new KeyManager[0];
+        });
+        kernel.getContext().put(SecurityService.class, securityService);
```
jbutler, line 296: Maybe I'm missing something. Isn't there a race if you don't store this in the context before starting nucleus?

jbutler: Oh, ok, so you're essentially hot patching and then relying on an implementation detail (that the security service is retrieved from the context rather than cached).

Can we inject this mock up front and then update the response? I think that achieves the same but without relying on that implementation detail in Nucleus

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,63 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        // Once the request is processed, the sync queue should be empty.
+        assertEmptySyncQueue(RealTimeSyncStrategy.class);
+    }
+
+    class FakeIotDataPlaneClientWrapper extends IotDataPlaneClientWrapper {
+        IotDataPlaneClientFactory factory;
+
+        public FakeIotDataPlaneClientWrapper(IotDataPlaneClientFactory iotDataPlaneClientFactory) {
+            super(iotDataPlaneClientFactory);
+            factory = iotDataPlaneClientFactory;
```
Nelsonochoam, line 314:
```suggestion
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,63 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        }
+
+        @Override
+        public GetThingShadowResponse getThingShadow(String thingName, String shadowName)
+                throws IoTDataPlaneClientCreationException {
+
+            factory.getIotDataPlaneClient(); // we're only interested in getting a good client in this fake wrapper.
```
Nelsonochoam, line 321:
```suggestion
            iotDataPlaneClientFactory.getIotDataPlaneClient(); // we're only interested in getting a good client in this fake wrapper.
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
Nelsonochoam: The parent is already storing the factory on a variable

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,63 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+            throws InterruptedException, TLSAuthException, IoTDataPlaneClientCreationException {
+        ignoreExceptionOfType(context, TLSAuthException.class);
+        ignoreExceptionOfType(context, RetryableException.class);
+
+        SecurityService securityService = mock(SecurityService.class);
+        when(securityService.getDeviceIdentityKeyManagers()).thenThrow(TLSAuthException.class);
+        kernel.getContext().put(SecurityService.class, securityService);
+
+        IotDataPlaneClientFactory factory = kernel.getContext().get(IotDataPlaneClientFactory.class);
+        IotDataPlaneClientWrapper wrapper = spy(new FakeIotDataPlaneClientWrapper(factory));
```
Nelsonochoam, line 279: Instead of the full blown class below since we just overriding a single method we could

```java
spy(new IotDataPlaneClientWrapper(factory) {
      @Override
      public GetThingShadowResponse getThingShadow(String thingName, String shadowName) throws IoTDataPlaneClientCreationException {
         iotDataPlaneClientFactory.getIotDataPlaneClient(); // throws an IoTDataPlaneClientCreationException given security service is not ready
         return GetThingShadowResponse.builder().build();
     }
});
```

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +264,63 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+        // creation fails. If the data plane client creation succeeds, the request is processed once again.
+        CountDownLatch cdl = new CountDownLatch(1);
+        reset(securityService);
+        when(securityService.getDeviceIdentityKeyManagers()).thenAnswer((invocation) -> {
+            cdl.countDown();
+            return new KeyManager[0];
+        });
+        assertThat("request is retried with a new client", cdl.await(10, TimeUnit.SECONDS), is(true));
+        verify(securityService, atMostOnce()).getDeviceIdentityKeyManagers();
+        verify(wrapper, timeout(5000).atLeast(2)).getThingShadow("Thing1", "");
```
jbutler, line 304: Ok, last comment. This test is very tightly coupled with implementation. You should not assert that `getDeviceIdentityKeyManagers` is called. Assume Nucleus will fix this at some point, at which point Shadow won't need this workaround. If we change the implementation, we should not need to update this test.

All you want to test for here is that the shadow operations are retried. Everything else is an implementation detail.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
saranyailla: `iotDataPlaneClientFactory` is a private field in the wrapper. So, I added this to use the injected factory.

src/integrationtests/java/com/aws/greengrass/integrationtests/ShadowManagerTest.java
```diff
@@ -247,4 +263,48 @@ void GIVEN_shadow_manager_WHEN_strategy_config_resets_THEN_respond_to_config_upd
...
+
+        SecurityService securityService = mock(SecurityService.class);
+        when(securityService.getDeviceIdentityKeyManagers()).thenThrow(TLSAuthException.class);
+        kernel.getContext().put(SecurityService.class, securityService);
+
+        IotDataPlaneClientFactory factory = kernel.getContext().get(IotDataPlaneClientFactory.class);
+        IotDataPlaneClientWrapper wrapper = spy(new IotDataPlaneClientWrapper(factory) {
+            @Override
+            public GetThingShadowResponse getThingShadow(String thingName, String shadowName) throws IoTDataPlaneClientCreationException {
+                factory.getIotDataPlaneClient(); // throws an IoTDataPlaneClientCreationException if security service is not ready
```
Nelsonochoam, line 281: Just to avoid referencing the external factory variable
```suggestion
                this.iotDataPlaneClientFactory.getIotDataPlaneClient(); // throws an IoTDataPlaneClientCreationException if security service is not ready
```

jbutler (approved):

src/test/java/com/aws/greengrass/shadowmanager/sync/IotDataPlaneClientWrapperTest.java
```diff
@@ -217,4 +219,28 @@ void GIVEN_exception_during_get_WHEN_get_thing_shadow_THEN_throw_sdk_exception(C
...
+    void GIVEN_invalid_client_creation_WHEN_update_thing_shadow_THEN_throw_IoTDataPlaneClientCreationException() throws IoTDataPlaneClientCreationException {
+        reset(iotDataPlaneClientFactory);
+        when(iotDataPlaneClientFactory.getIotDataPlaneClient()).thenThrow(IoTDataPlaneClientCreationException.class);
+        IotDataPlaneClientWrapper iotDataPlaneClientWrapper = new IotDataPlaneClientWrapper(iotDataPlaneClientFactory);
+        assertThrows(IoTDataPlaneClientCreationException.class, () -> iotDataPlaneClientWrapper.updateThingShadow(THING_NAME, SHADOW_NAME,"".getBytes(StandardCharsets.UTF_8)));
+    }
+
+    @Test
+    void GIVEN_invalid_client_creation_WHEN_delete_thing_shadow_THEN_throw_IoTDataPlaneClientCreationException() throws IoTDataPlaneClientCreationException {
+        reset(iotDataPlaneClientFactory);
```
MikeDombo, line 241: nit: put this in the `before` if we have to do it all the time?

MikeDombo (approved):

PR #143, opened 2022-12-21
refactor: shadow doc size config updates

saranyailla:
**Issue #, if available:**

**Description of changes:**
Refactor handling shadow doc size configuration updates to clean up shadow manager service class. 

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -19,6 +19,7 @@ import com.aws.greengrass.mqttclient.CallbackEventManager;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.shadowmanager.configuration.ComponentConfiguration;
 import com.aws.greengrass.shadowmanager.configuration.RateLimitsConfiguration;
+import com.aws.greengrass.shadowmanager.configuration.ShadowDocSizeConfiguration;
 import com.aws.greengrass.shadowmanager.exception.InvalidConfigurationException;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowIPCHandler;
 import com.aws.greengrass.shadowmanager.ipc.DeleteThingShadowRequestHandler;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -320,23 +316,8 @@ private void configureSynchronization(Node newv) {
...
-            Validator.setMaxShadowDocumentSize(newMaxShadowSize);
-            logger.atDebug()
-                    .setEventType("config")
-                    .kv("maxShadowSize", newMaxShadowSize)
-                    .log();
-        } catch (InvalidConfigurationException e) {
-            serviceErrored(e);
-        }
+    private void configureShadowDocSize(ShadowDocSizeConfiguration shadowDocSizeConfiguration) {
+        Validator.setMaxShadowDocumentSize(shadowDocSizeConfiguration.getMaxShadowDocSizeConfiguration());
```
Nelsonochoam, line 320: I know you didn't do this but this kind of validator object is less than ideal and not really object oriented. It has a mix of concerns and business logic that crosses boundaries... Ideally we had a Shadow class locally that did this... I think it is ok for now but we should look into fixing this.

The `Validator` is basically a "service object", those are generally bad because they break encapsulation. Here is nice article that talks about it [Why Service Objects Are a Bad Idea](https://www.codewithjason.com/rails-service-objects/?utm_source=pocket_reader)

> They’re chunks of procedural code masquerading as object-oriented code.

Nelsonochoam: Maybe add a TODO to fix this and stop using that class

src/main/java/com/aws/greengrass/shadowmanager/configuration/ShadowDocSizeConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+     * @return rate limits configuration objects
+     */
+    public static ShadowDocSizeConfiguration from(Topics serviceTopics) {
+        return new ShadowDocSizeConfiguration(getMaxShadowDocSizeFromConfig(serviceTopics));
+    }
+
+    private static int getMaxShadowDocSizeFromConfig(Topics topics) {
+        int newMaxShadowSize = Coerce.toInt(
+                topics.findOrDefault(DEFAULT_DOCUMENT_SIZE, CONFIGURATION_MAX_DOC_SIZE_LIMIT_B_TOPIC));
+        Validator.validateMaxShadowSize(newMaxShadowSize);
```
Nelsonochoam, line 37: Let's move the logic from the `Validator. validateMaxShadowSize` to a private method on this class. I explained in another comment why the validator class might not be well suited.

Nelsonochoam (approved):

MikeDombo (approved):

PR #163, opened 2023-01-17
fix: update all core thing sync configurations when the core thing na…

saranyailla:
…me changes

**Issue #, if available:**

**Description of changes:**
When the core thing name changes in the device configuration, we want all of the shadow sync configurations of the core thing to be updated.
For eg, for the core thing named `coreThing`, 3 shadows are configured in the shadow manager component as follows.

- coreThing-
- coreThing-nameShadow1
- coreThing-nameShadow2

Now, when the core thing name is updated to `newCoreThing`, we want the shadows to be updated as follows
- newCoreThing-
- newCoreThing-nameShadow1
- newCoreThing-nameShadow2
But that's not the case as we're updating only one of the above shadows as seen [here](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/9bc5c2495003cdc51a87efa8f027abdd0f1ed136/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L449). 

This change in the PR updates all the shadow sync configurations of a coreThing when it changes. 

Added an integration test to verify the same. 

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -64,6 +64,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
@@ -210,7 +211,7 @@ public class ShadowManager extends PluginService {
             // First time.
             return;
         }
-        getCoreThingShadowSyncConfiguration(oldThingName).ifPresent(thingShadowSyncConfiguration ->
+        getCoreThingShadowSyncConfiguration(oldThingName).forEach(thingShadowSyncConfiguration ->
                 thingShadowSyncConfiguration.setThingName(thingName));
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -442,11 +443,11 @@ private void insertSyncInfoIfNotPresent(long epochSeconds, ThingShadowSyncConfig
                 .build());
     }
 
-    private Optional<ThingShadowSyncConfiguration> getCoreThingShadowSyncConfiguration(String thingName) {
+    private List<ThingShadowSyncConfiguration> getCoreThingShadowSyncConfiguration(String thingName) {
```
rbattle, line 446: will this ever be a list that is not either length 0 or 1? why did you have to change it from optional to list?

saranyailla: In the current code, we're using the [same method](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/main/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L213-L214) while updating the thing shadow sync configuration changes. So, we need a list instead of optional to update all the shadows. Description of the PR has better context with an example.

nikkhilmuthye (approved):

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
Nelsonochoam, line 446:
```suggestion
    private List<ThingShadowSyncConfiguration> getCoreThingShadowSyncConfigurations(String thingName) {
```

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -297,12 +298,12 @@ private void configureSynchronization(Node newv) {
             this.syncHandler.setSyncConfigurations(this.syncConfiguration.getSyncConfigurations());
 
             // Subscribe to the thing name topic if the Nucleus thing shadows have been synced.
-            Optional<ThingShadowSyncConfiguration> coreThingConfig =
+            List<ThingShadowSyncConfiguration> coreThingConfig =
```
Nelsonochoam, line 301:
```suggestion
            List<ThingShadowSyncConfiguration> coreThingConfigs =
```

Nelsonochoam (approved):
Left some minor comments but looks good

PR #166, opened 2023-01-30
fix: synchronize access to sync information

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Shadow manager maintains `sync` information in its database, to help make decisions on when a shadow should be synced to the cloud.  Currently, this table is read and (sometimes) written by multiple threads:
1) MqttClient thread when receiving updates from IoT Core
2) IPC thread when receiving local shadow update,
3) Sync thread, when a cloud or local update sync request is executing.  

This change ensures that read/write access to sync information is synchronized.

A customer has experienced a race condition in the local --> cloud sync flow where updates made in quick succession result in certain sync requests being dropped, leaving the cloud shadow in the wrong state.  Logs show that the IPC thread that receives the local update wrongfully determines that no update is needed, based on outdated information from the sync table. 

```
2023-01-26T09:08:35.278Z [INFO] (IPC THREAD) com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl: Updating sync info. {thing name=thing, shadow name=shadow, cloud-version=1, local-version=4}
2023-01-26T09:08:35.357Z [INFO] (SYNC THREAD) com.aws.greengrass.shadowmanager.ShadowManagerDAOImpl: Updating sync info. {thing name=thing, shadow name=shadow, cloud-version=2, local-version=3}
```

* Integration tests
* Checked that there was no unit/integ regression.

**Checklist:**
 

src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
```diff
@@ -172,21 +172,25 @@ public void execute(SyncContext context) throws RetryableException, SkipSyncRequ
...
     public boolean isUpdateNecessary(SyncContext context) throws SkipSyncRequestException, UnknownShadowException {
-        Optional<ShadowDocument> shadowDocument = context.getDao().getShadowThing(getThingName(), getShadowName());
-
-        //TODO: store this information in a return object to avoid unnecessary calls to DAO.
-        SyncInformation currentSyncInformation = context.getDao()
-                .getShadowSyncInformation(getThingName(), getShadowName())
-                .orElseThrow(() -> new UnknownShadowException("Shadow not found in sync table"));
-
-        return isUpdateNecessary(shadowDocument, currentSyncInformation, context);
+        // This method is called in context of an IPC-handling thread.
```
nikkhilmuthye, line 182: Where is this method called from the IPC handling thread?

jcosentino11: here's the sequence of calls:

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java#L253

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/sync/SyncHandler.java#L246

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java#L280

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java#L189

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java#L212

nikkhilmuthye: The `BaseSyncStrategy` should happen in a separate thread. It should not be a sequential call since `putSyncRequest` just puts the request in the sync queue. We have a separate function which executes these sync requests.

nikkhilmuthye: Spoke offline. Change looks good. Add an integ test to verify this change.

rbattle (commented):
This looks fine to me - I had to go through the code again to see that the other form of `isUpdateNecessary` is called when the sync queue is executing the request.

This check was added ind #106 . Will it cause a regression for that case and cause unnecessary full syncs?

nikkhilmuthye:
> This looks fine to me - I had to go through the code again to see that the other form of `isUpdateNecessary` is called when the sync queue is executing the request.
> 
> This check was added in #106 . Will it cause a regression for that case and cause unnecessary full syncs? That was for a local update case, but I want to be certain that it won't cause an issue

Ah yes. That would cause a full sync. When a local update happens, we update the cloud shadow. Once the cloud shadow updates, we get an update on the MQTT topic that the device is subscribed to. So, the cloud update requests will get added to the queue. If the device updates the shadow during that time, these two requests will merge into a full sync request which is unnecessary since we only need to update the cloud shadow.

src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
nikkhilmuthye: I believe we need to block on the shadow name. So have a central lock system based on the shadow name.

rbattle: right. it would block on the shadow. I wasn't thinking of a global lock

rbattle: for the case where you a syncing a single shadow with constant updates, it _could_ slow that down, but it may be a better overall solution

nikkhilmuthye: Right. A central locking system based on the shadow name might delay it by a few ms. But I also think that its a better overall solution.

src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
```diff
@@ -86,86 +86,88 @@ public void merge(CloudUpdateSyncRequest other) {
...
+                        JsonUtil.getPayloadBytes(updateDocument));
+                cloudUpdatedVersion = getUpdatedVersion(response.payload().asByteArray()).orElse(cloudVersion + 1);
+                logger.atDebug()
+                        .kv(LOG_THING_NAME_KEY, getThingName())
+                        .kv(LOG_SHADOW_NAME_KEY, getShadowName())
+                        .kv(LOG_LOCAL_VERSION_KEY, currentSyncInformation.getLocalVersion())
+                        .kv(LOG_CLOUD_VERSION_KEY, currentSyncInformation.getCloudVersion())
+                        .kv(LOG_UPDATED_CLOUD_VERSION_KEY, cloudUpdatedVersion)
+                        .log("Successfully updated cloud shadow document");
+            } catch (ConflictException e) {  // NOPMD - Throw ConflictException instead of treated as SdkServiceException
```
MikeDombo, line 121: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: An exception is caught and re-thrown within the catch clause without any other operations being performed on the exception, making the catch clause unnecessary.

**Fix**: Perform additional operations on the exception that has been caught. You can wrap a standard exception into a custom exception to add additional information or special handling. When you do that, make sure to set the original exception as the cause. Otherwise, you will lose the stack trace and message of the original exception, which will make it difficult to analyze the event that caused the exception.  
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

src/main/java/com/aws/greengrass/shadowmanager/sync/model/CloudUpdateSyncRequest.java
```diff
@@ -86,86 +86,88 @@ public void merge(CloudUpdateSyncRequest other) {
     public void execute(SyncContext context) throws RetryableException, SkipSyncRequestException,
             ConflictException, UnknownShadowException, InterruptedException {
 
-        Optional<ShadowDocument> shadowDocument = context.getDao().getShadowThing(getThingName(), getShadowName());
-
-        //TODO: store this information in a return object to avoid unnecessary calls to DAO.
-        SyncInformation currentSyncInformation = context.getDao()
-                .getShadowSyncInformation(getThingName(), getShadowName())
-                .orElseThrow(() -> new UnknownShadowException("Shadow not found in sync table"));
+        synchronized (context.getSynchronizeHelper().getThingShadowLock(this)) {
```
rbattle, line 89: do you also need to synchronize in `isUpdateNecessary(SyncContext context)` ?

jcosentino11: technically no but it certainly wouldn't hurt.  that method is only ever called within this synchronize block https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/main/src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java#L111.

jcosentino11: just saw the edit about adding a comment, will do

jcosentino11: updated:  when answering the question of "why not needed for other types of requests" i think the same issue would happen for LocalUpdateSyncRequest, since `isUpdateNecessary` which reads/writes to sync table, is called in context of mqtt callback thread and the sync thread

jcosentino11: however if we need to have similar synchronization for LocalUpdateSyncRequest, there's the issue where we can't block the subscribe callback thread in https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/8e24c7de2f5d091fb6a42247742d28be76f96a5a/src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java#L279

jcosentino11: i guess we could dispatch an async task, but now things are becoming more complex

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -432,6 +436,71 @@ void GIVEN_synced_shadow_WHEN_local_update_THEN_cloud_updates(Class<?extends Bas
                 any(software.amazon.awssdk.services.iotdataplane.model.UpdateThingShadowRequest.class));
     }
 
+    @ParameterizedTest
+    @ValueSource(classes = {RealTimeSyncStrategy.class, PeriodicSyncStrategy.class})
+    void GIVEN_synced_shadow_WHEN_multiple_local_updates_THEN_cloud_updates(Class<?extends BaseSyncStrategy> clazz, ExtensionContext context) throws IOException,
```
nikkhilmuthye, line 441: Can we add a test for LocalShadowUpdate as well?

Also, can you check if you can intertwine a Cloud update and a local update in a similar manner in another test?

jcosentino11: added

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -283,7 +284,22 @@ private void handleUpdate(MqttMessage message) {
         String shadowName = shadowRequest.getShadowName();
         logger.atDebug().kv(LOG_THING_NAME_KEY, thingName).kv(LOG_SHADOW_NAME_KEY, shadowName)
                 .log("Received cloud update sync request");
-        syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload());
+        CompletableFuture
```
nikkhilmuthye, line 287: Why are you using `CompletableFuture` here? Why not use `ExecutorService`?

src/test/java/com/aws/greengrass/shadowmanager/sync/model/LocalUpdateSyncRequestTest.java
```diff
@@ -91,7 +92,7 @@ class LocalUpdateSyncRequestTest {
     @BeforeEach
     void setup() throws IOException {
         syncContext = new SyncContext(mockDao, mockUpdateThingShadowRequestHandler,
-                mock(DeleteThingShadowRequestHandler.class), mock(IotDataPlaneClientWrapper.class));
+                mock(DeleteThingShadowRequestHandler.class), mock(IotDataPlaneClientWrapper.class), new ShadowWriteSynchronizeHelper());
```
nikkhilmuthye, line 95: Why not use a mock?

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
jcosentino11: Mostly for ease of logging if an exception happens within the future (the `whenComplete()` call below), but it does run in context of ExecutorService: `.runAsync(() -> blah, executorService)`

src/test/java/com/aws/greengrass/shadowmanager/sync/model/LocalUpdateSyncRequestTest.java
jcosentino11: since its just a map of locks, wouldn't mocking be more complicated to achieve the same functionality?

jcosentino11: wanted to exercise locking within sync testing which is why using real class here and cloud update test, while using mocks everywhere else

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
rbattle: what exception can occur in `pushLocalUpdateSyncRequest`?

Is there a chance of local update requests being pushed out of order?

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -283,7 +284,22 @@ private void handleUpdate(MqttMessage message) {
...
+                // There is a small chance of blocking when a thing/shadow lock is obtained
+                // within this call, because the lock also surrounds an IoT dataplane call
+                // in CloudUpdateSyncRequest.
+                .runAsync(() -> syncHandler.pushLocalUpdateSyncRequest(thingName, shadowName, message.getPayload()),
+                        executorService)
+                .whenComplete((unused, e) -> {
+                    if (e != null) {
+                        logger.atError().cause(e)
+                                .kv("thingName", thingName)
+                                .kv("shadowName", shadowName)
```
rbattle, line 299: nit: LOG_THING_NAME_KEY and LOG_SHADOW_NAME_KEY

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
jcosentino11: > what exception can occur in `pushLocalUpdateSyncRequest`?

in LocalUdpateSyncRequest#isUpdateNecessary:

* `throw new SkipSyncRequestException(e);` for malformed shadow payload 
* new UnknownShadowException("Shadow not found in sync table") when searching in sync table


> Is there a chance of local update requests being pushed out of order?

presumably, since mqtt doesn't have ordering guarantees, or if we receive the message twice since it's QOS1

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
jcosentino11: fixed

rbattle (approved):

nikkhilmuthye (approved):

PR #167, opened 2023-02-07
fix: handle out-of-order cloud updates

jcosentino11:
**Issue #, if available:**

**Description of changes:**

When cloud updates are received over MQTT, there is a (small) chance that we can receive updates out-of-order.  Also, with the proposal of https://github.com/aws-greengrass/aws-greengrass-shadow-manager/pull/166, cloud updates would be processed asynchronously, further increasing the chances of out-of-order processing.  

Functionally, when we receive out-of-order cloud updates (say, B then A), this is automatically detected and a ConflictError is thrown.  This causes a full sync, which puts the local shadow in sync with the cloud.

The first issue however, is when we receive the other out of order message (A, from the above example) after the conflict has been dealt with.  This update has the incorrect, previous cloud version, and will be written to the sync table.  Effectively, this will cause unnecessary full syncs, as more conflicts will be detected due to the discrepancy in cloud versions.

This change proposes that updates to the sync table will only be committed for new versions when a cloud update is received.

A second issue is when the out of order messages are merged together. Currently, the old version is merged into the new version, causing the merged document to have the older version.  This change addresses this by comparing versions before selecting merge order.

See above description

New integration tests

**Checklist:**
 
### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
@@ -39,6 +39,7 @@ import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.slf4j.event.Level;
+import software.amazon.awssdk.aws.greengrass.model.ConflictError;
 import software.amazon.awssdk.aws.greengrass.model.DeleteThingShadowRequest;
 import software.amazon.awssdk.aws.greengrass.model.UpdateThingShadowRequest;
 import software.amazon.awssdk.core.SdkBytes;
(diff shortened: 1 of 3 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
```diff
@@ -728,6 +729,196 @@ void GIVEN_cloud_update_request_WHEN_retryable_thrown_AND_new_cloud_update_reque
...
+        assertThat("local version", syncInfo.get().get().getLocalVersion(), is(1L));
+
+        SyncHandler syncHandler = kernel.getContext().get(SyncHandler.class);
+
+        when(iotDataPlaneClientFactory.getIotDataPlaneClient().getThingShadow(any(GetThingShadowRequest.class))).thenReturn(finalCloudStateShadowResponse);
+
+        // receive cloud update 2 of 2 (out of order)
+        syncHandler.pushLocalUpdateSyncRequest(MOCK_THING_NAME_1, CLASSIC_SHADOW, JsonUtil.getPayloadBytes(cloudUpdateDocument2));
+        // receive cloud update 1 of 2 (out of order)
+        syncHandler.pushLocalUpdateSyncRequest(MOCK_THING_NAME_1, CLASSIC_SHADOW, JsonUtil.getPayloadBytes(cloudUpdateDocument));
```
rbattle, line 842: will this consistently merge or could it process the first sync request thread before the next one is added?

src/main/java/com/aws/greengrass/shadowmanager/sync/model/LocalUpdateSyncRequest.java
```diff
@@ -68,17 +68,30 @@ public LocalUpdateSyncRequest(String thingName,
...
         }
-        JsonMerger.merge(oldValueJson.get(), newValueJson.get());
-        updateDocument = JsonUtil.getPayloadBytes(oldValueJson.get());
+
+        JsonNode oldValueJson = currDocJson.get();
+        JsonNode newValueJson = updateDocJson.get();
+
+        if (JsonUtil.hasVersion(oldValueJson) && JsonUtil.hasVersion(newValueJson)
+                && JsonUtil.getVersion(oldValueJson) > JsonUtil.getVersion(newValueJson)) {
+            // updates received out of order,
```
rbattle, line 87: can you add a debug log here saying it received version X after receiving version X+1

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
jcosentino11: it'll consistently merge for periodic sync strategy, but not for realtime.  will look into making more reliable

src/main/java/com/aws/greengrass/shadowmanager/sync/model/LocalUpdateSyncRequest.java
jcosentino11: done

src/integrationtests/java/com/aws/greengrass/integrationtests/SyncTest.java
jcosentino11: fixed

rbattle (dismissed):

src/main/java/com/aws/greengrass/shadowmanager/sync/model/LocalUpdateSyncRequest.java
```diff
@@ -68,17 +68,32 @@ public LocalUpdateSyncRequest(String thingName,
...
+        if (!updateDocJson.isPresent()) {
             return;
         }
-        JsonMerger.merge(oldValueJson.get(), newValueJson.get());
-        updateDocument = JsonUtil.getPayloadBytes(oldValueJson.get());
+
+        JsonNode oldValueJson = currDocJson.get();
+        JsonNode newValueJson = updateDocJson.get();
+
+        if (JsonUtil.hasVersion(oldValueJson) && JsonUtil.hasVersion(newValueJson)
```
nikkhilmuthye, line 85: I feel this is a good start. But we might want to consider a per node update instead of a per version update in the future. It is a bit more work. But that will ensure that the data is correct.

nikkhilmuthye (dismissed):

nikkhilmuthye (approved):

rbattle (approved):

PR #170, opened 2023-02-15
fix: respect sync direction when queueing full sync on conflict

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Changes the type of request enqueued when a conflict is detected from a `FullShadowSyncRequest`, to one of `OverwriteCloudShadowRequest`/`OverwriteLocalShadowRequest`/`FullShadowSyncRequest`, based on the sync direction.

Also refactors so that sync direction is wrapped in a class which is injected wherever needed.  This means that all usages will have the same reference to direction, rather than needing to ensure setters are called.

To prevent full syncs from happening when a conflict is detected, when sync direction is not BETWEEN_DEVICE_AND_CLOUD.

new unit tests

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
@@ -12,7 +12,10 @@ import com.aws.greengrass.shadowmanager.exception.SkipSyncRequestException;
 import com.aws.greengrass.shadowmanager.exception.UnknownShadowException;
 import com.aws.greengrass.shadowmanager.sync.RequestBlockingQueue;
 import com.aws.greengrass.shadowmanager.sync.Retryer;
+import com.aws.greengrass.shadowmanager.sync.model.DirectionWrapper;
 import com.aws.greengrass.shadowmanager.sync.model.FullShadowSyncRequest;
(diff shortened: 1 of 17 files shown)


src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java
```diff
@@ -410,15 +418,17 @@ protected void syncLoop() {
                             .addKeyValue(LOG_SHADOW_NAME_KEY, currProcessingShadowName)
                             .log("Received conflict when processing request. Retrying as a full sync");
                     // add back to queue to merge over any shadow request that came in while it was executing
-                    request = syncQueue.offerAndTake(new FullShadowSyncRequest(currProcessingThingName,
-                            currProcessingShadowName), true);
+                    request = syncQueue.offerAndTake(
+                            fullSyncRequestBasedOnDirection(currProcessingThingName, currProcessingShadowName),
```
jcosentino11, line 422: this and the other change below are the fix, the rest is refactor

MikeDombo (approved):
LGTM. Is this addressing the race conditions though, or is this something else?

jcosentino11:
> LGTM. Is this addressing the race conditions though, or is this something else?

Not directly related, but preventing full syncs when we don't expect them will help when reasoning about the race condition fix

saranyailla (approved):

PR #172, opened 2023-03-07
feat: run uats against release candidate

Nelsonochoam:
**Description of changes:**
Modify the uat setup so that the tests run against the current version of the code and not against the artifact already deployed in production.

Required to effectively test uats on our release pipeline

Ran the uats


uat/testing-features/src/main/java/com/aws/greengrass/ShadowSteps.java
```diff
@@ -112,6 +112,7 @@ private String randomName() {
         return String.format("e2e-%d-%s", System.currentTimeMillis(), UUID.randomUUID().toString());
     }
 
+    @SuppressWarnings("PMD.UnusedFormalParameter")
```
Nelsonochoam, line 115: There as a PMD violation `version` param is not being used

saranyailla (approved):

dgggIoT (approved):

PR #177, opened 2023-03-24
fix: wait for DB write operations to finish

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Use a separate thread pool for DB write operations to avoid interrupting the threads during nucleus shutdown process.

To improve DB durability in case of power off, we're adding the following. More info [here](http://www.h2database.com/html/advanced.html#:~:text=In%20H2%2C%20after,that%20support%20it.)
- Use [CHECKPOINT_SYNC](https://www.h2database.com/html/commands.html#checkpoint_sync) to flush the data to disk and force all system buffers be written to the underlying device.
- Set [WRITE_DELAY](https://www.h2database.com/html/commands.html#set_write_delay) to 0ms to reduce the delay between commit and log flush. 

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
@@ -11,12 +11,8 @@ import com.aws.greengrass.util.Pair;
 
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.Semaphore;
 
 public interface ShadowManagerDAO {
-    int maxPermits = 10;
-    Semaphore dbWriteOperations = new Semaphore(maxPermits);
-
     /**
      * Attempts to obtain a shadow document from the local shadow storage.
      *
@@ -114,6 +110,4 @@ public interface ShadowManagerDAO {
      * @return true if the insert is successful; Else false.
      */
     boolean insertSyncInfoIfNotExists(SyncInformation request);
-
-    void waitForDBOperationsToFinish();
 }
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
@@ -11,12 +11,8 @@ import com.aws.greengrass.util.Pair;
 
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 
 public interface ShadowManagerDAO {
-    ExecutorService dbWriteThreadPool = Executors.newCachedThreadPool();
-
     /**
      * Attempts to obtain a shadow document from the local shadow storage.
      *
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
@@ -11,8 +11,12 @@ import com.aws.greengrass.util.Pair;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.Semaphore;
 
 public interface ShadowManagerDAO {
+    int maxPermits = 10;
+    Semaphore dbWriteOperations = new Semaphore(maxPermits);
+
     /**
      * Attempts to obtain a shadow document from the local shadow storage.
      *
@@ -110,4 +114,6 @@ public interface ShadowManagerDAO {
      * @return true if the insert is successful; Else false.
      */
     boolean insertSyncInfoIfNotExists(SyncInformation request);
+
+    void waitForDBOperationsToFinish();
 }
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
@@ -11,8 +11,12 @@ import com.aws.greengrass.util.Pair;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public interface ShadowManagerDAO {
+    ExecutorService dbWriteThreadPool = Executors.newCachedThreadPool();
+
     /**
      * Attempts to obtain a shadow document from the local shadow storage.
      *
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
@@ -18,7 +18,6 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.sql.Statement;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
@@ -368,21 +367,10 @@ public class ShadowManagerDAOImpl implements ShadowManagerDAO {
         }
     }
 
-    private void checkpointSync() {
-        try (Connection c = database.getPool().getConnection();
-             Statement statement = c.createStatement()) {
-             statement.execute("CHECKPOINT SYNC");
-        } catch (SQLException | IllegalStateException e) {
-            throw new ShadowManagerDataException(e);
-        }
-    }
-
     private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
         try {
-            T result = database.getDbWriteThreadPool().submit(() ->
+            return database.getDbWriteThreadPool().submit(() ->
                     execute(sql, thunk)).get(10, TimeUnit.SECONDS);
-            checkpointSync();
-            return result;
         } catch (InterruptedException | ExecutionException | TimeoutException e) {
             throw new ShadowManagerDataException(e);
         }
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
@@ -87,10 +87,11 @@ public class ShadowManagerDAOImpl implements ShadowManagerDAO {
                 .kv(LOG_THING_NAME_KEY, thingName)
                 .kv(LOG_SHADOW_NAME_KEY, shadowName)
                 .log("Deleting shadow");
+        String sql = "UPDATE documents SET deleted = 1, document = null, updateTime = ?, version = ?"
+                + " WHERE thingName = ? AND shadowName = ?";
         return getShadowThing(thingName, shadowName)
                 .flatMap(shadowDocument ->
-                        execute("UPDATE documents SET deleted = 1, document = null, updateTime = ?, version = ?"
-                                        + " WHERE thingName = ? AND shadowName = ?",
+                        executeWriteOperation(sql,
                                 preparedStatement -> {
                                     preparedStatement.setLong(1, Instant.now().getEpochSecond());
                                     preparedStatement.setLong(2, shadowDocument.getVersion() + 1);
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +369,16 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
             throw new ShadowManagerDataException(e);
         }
     }
+
+    private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
+        try {
+            dbWriteOperations.acquire();
+        } catch (InterruptedException e) {
+            logger.atDebug().log("Interrupted before performing the DB operation");
+            Thread.currentThread().interrupt();
```
MikeDombo, line 378: exit here, you were interrupted, so you shouldn't then try to write.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +369,16 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
...
+        }
+        T result = execute(sql, thunk);
+        dbWriteOperations.release();
```
MikeDombo, line 381: use try-finally to guarantee release.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
```diff
@@ -11,8 +11,12 @@
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.Semaphore;
 
 public interface ShadowManagerDAO {
+    int maxPermits = 10;
```
rbattle, line 17: nit: might need to increase this - we have some configuration for number of operations per second. May want to just use that (even if that includes read operations)

saranyailla: Should not this be the number of threads operating on DB (for write) at any given time? Currently, we have three threads that can do DB write (IPC thread, Sync and service thread), so I had the max limit as 10.

MikeDombo (commented):
Update the PR description.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -352,6 +355,16 @@ public boolean insertSyncInfoIfNotExists(SyncInformation request) {
 
     }
 
+    @Override
+    public void waitForDBOperationsToFinish() {
```
Nelsonochoam, line 359: How was this tested? - We should be able to write an integration test for this.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -352,6 +355,16 @@ public boolean insertSyncInfoIfNotExists(SyncInformation request) {
...
+        if (!dbWriteOperations.tryAcquire(maxPermits)) {
```
Nelsonochoam, line 360: Why call this and then if it fails to acquire we wait call the blocking version?. We want to basically ensure we grab all the permits right?

```java
logger.atDebug().log("Waiting for the DB write operations to finish");
dbWriteOperations.acquireUninterruptibly(maxPermits);
```

Also this method partially works, we are acquiring all the permits and immediately releasing them there is no guarantee that another thread can grab a permit right after we release them that might lead to corruption. (I know you are calling this on the release method but given there are not tests for this someone in the future might just see this and remove it given they might not have the context.


Is there a way we could reject all write requests after this is called?  Maybe reducing the permits to 0?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
Nelsonochoam: We might want to add a comment as to why that number is 10. I am pretty sure someone would see this in the future and don't know why it is 10 rather than 2 or 100. Is there a way to dynamically set the value based on the number of threads being used to access the DAO?

What if the number of threads gets increased to something higher than 10?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
saranyailla: Calling this code from the db close method now.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
rbattle: right - for some reason I thought there would be a thread per request from IPC

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +363,17 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
             throw new ShadowManagerDataException(e);
         }
     }
+
+    private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
+        try {
+            database.getDbWriteOperations().acquire();
+            return execute(sql, thunk);
+        } catch (InterruptedException e) {
+            logger.atDebug().log("Interrupted before performing the DB operation");
```
rbattle, line 372: do you need to reset the interrupt here: `Thread.currentThread().interrupt();`

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +366,13 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
             throw new ShadowManagerDataException(e);
         }
     }
+
+    private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
+        try {
+            return dbWriteThreadPool.submit(() -> execute(sql, thunk)).get(10, TimeUnit.SECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
```
MikeDombo, line 373: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: InterruptedException is ignored. This can delay thread shutdown and clear the thread’s interrupt status. Only code that implements a thread’s interruption policy can swallow an interruption request.

**Fix**: Rethrow the InterruptedException or reinterrupt the current thread using *Thread.currentThread().interrupt()* so that higher-level interrupt handlers can function correctly.
 If you are wrapping the InterruptedException inside a RuntimeException, call *Thread.currentThread().interrupt()* before throwing the RuntimeException.

Learn more about [interrupts](https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html) and [dealing with InterruptedException](https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html?ca=drs-#N10187)

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAO.java
```diff
@@ -11,8 +11,12 @@
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public interface ShadowManagerDAO {
+    ExecutorService dbWriteThreadPool = Executors.newCachedThreadPool();
```
MikeDombo, line 18: call shutdown, but not shutdownNow when closing shadow manager. This will prevent new threads from spawning.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
```diff
@@ -41,6 +41,7 @@ public class ShadowManagerDatabase implements Closeable {
             + ";RETENTION_TIME=1000" // ms - time to keep values for before writing to disk (default is 45000)
             + ";DEFRAG_ALWAYS=TRUE" // defragment db on shutdown (ensures only a single value in db on close)
             + ";COMPRESS=TRUE" // compress large objects (clob/blob) (default false)
+            + ";WRITE_DELAY=0" // maximum delay between commit and flushing the log
```
MikeDombo, line 44: what's the difference with retention time? They seem really similar.

MikeDombo: and if we have this, do we also need the sql command?

saranyailla: The way I understood is that CHECKPOINT SYNC does more than just flushing the commit to log and that it flushes data and other index files as well.

saranyailla: IMO, if we're writing the log immediately, they we can let go of retention time as we no longer need it. If I remove it, it is set to default value of 45 seconds. Not confident about making it 0. Hence left it to what it was before.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +366,14 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
             throw new ShadowManagerDataException(e);
         }
     }
+
+    private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
+        try {
+            return database.getDbWriteThreadPool().submit(() ->
+                    execute(sql, thunk)).get(10, TimeUnit.SECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
```
jcosentino11, line 374: do we need to reset interrupt flag here? or if not i think a comment would be useful

src/test/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImplTest.java
```diff
@@ -80,6 +82,15 @@ class ShadowManagerDAOImplTest {
     @Captor
     private ArgumentCaptor<byte[]> bytesArgumentCaptor;
 
+    @BeforeEach
+    void setup() throws SQLException, IOException {
+        when(mockConnection.prepareStatement(anyString())).thenReturn(mockPreparedStatement);
+        lenient().when(mockDatabase.getDbWriteThreadPool()).thenReturn(Executors.newCachedThreadPool());
```
jcosentino11, line 88: shutdown thread pool after each?

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDAOImpl.java
```diff
@@ -360,4 +366,14 @@ private <T> T execute(String sql, SQLExecution<T> thunk) {
             throw new ShadowManagerDataException(e);
         }
     }
+
+    private <T> T executeWriteOperation(String sql, SQLExecution<T> thunk) {
+        try {
+            return database.getDbWriteThreadPool().submit(() ->
+                    execute(sql, thunk)).get(10, TimeUnit.SECONDS);
```
jcosentino11, line 373: thinking about edge case, is it possible for the execution to take longer than 10 seconds?  not sure what the default execution timeout is but we may want to explicitly set `setQueryTimeout` and then in theory we could just `get()` here

saranyailla: Added `setQueryTimeout` on statement but still keeping `get(timeout, seconds)` as we don't know when the thread actually starts executing the query and we can't wait indefinitely.

jcosentino11 (dismissed):

MikeDombo (approved):

jcosentino11 (approved):

PR #178, opened 2023-03-29
fix: drop reported delta in update shadow

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Specifying "delta" as part of an update request was allowed, but this does not work when reporting the shadow to IoT Core.

```
InvalidRequestException: State contains an invalid node: 'delta' (Service: IotDataPlane, Status Code: 400
```

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
@@ -43,6 +43,8 @@ import static com.aws.greengrass.shadowmanager.model.Constants.LOG_LOCAL_VERSION
 import static com.aws.greengrass.shadowmanager.model.Constants.LOG_SHADOW_NAME_KEY;
 import static com.aws.greengrass.shadowmanager.model.Constants.LOG_THING_NAME_KEY;
 import static com.aws.greengrass.shadowmanager.model.Constants.SHADOW_DOCUMENT_METADATA;
+import static com.aws.greengrass.shadowmanager.model.Constants.SHADOW_DOCUMENT_STATE;
+import static com.aws.greengrass.shadowmanager.model.Constants.SHADOW_DOCUMENT_STATE_DELTA;
 import static com.aws.greengrass.shadowmanager.util.JsonUtil.isNullOrMissing;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.UPDATE_THING_SHADOW;
 
@@ -129,6 +131,14 @@ public class UpdateThingShadowRequestHandler extends BaseRequestHandler {
                             .orElseThrow(() ->
                                     new InvalidRequestParametersException(ErrorMessage
                                             .createInvalidPayloadJsonMessage("")));
+
+                    // drop "delta" from the state (if we have a state).
+                    // delta isn't valid for users to set.
+                    if (updateDocumentRequest.has(SHADOW_DOCUMENT_STATE)
+                            && updateDocumentRequest.get(SHADOW_DOCUMENT_STATE).isObject()) {
+                        ((ObjectNode) updateDocumentRequest.get(SHADOW_DOCUMENT_STATE))
+                                .remove(SHADOW_DOCUMENT_STATE_DELTA);
+                    }
                     // Validate the payload schema
                     JsonUtil.validatePayloadSchema(updateDocumentRequest);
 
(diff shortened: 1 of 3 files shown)


Nelsonochoam (approved):

src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
```diff
@@ -129,6 +132,10 @@ public UpdateThingShadowHandlerResponse handleRequest(UpdateThingShadowRequest r
                             .orElseThrow(() ->
                                     new InvalidRequestParametersException(ErrorMessage
                                             .createInvalidPayloadJsonMessage("")));
+                    if (updateDocumentRequest.has(SHADOW_DOCUMENT_STATE)
```
nikkhilmuthye, line 135: I believe this would this be a breaking change for customers right? We don't process the `delta` node anyways. Why not just emit sending it to the cloud when we sync?

MikeDombo: it shouldn't break anyone. Anyone who is setting the delta today is already broken.

nikkhilmuthye: The local update shouldn't be broken though right? This will break the local shadow update as well. It will only break the customers who are syncing that shadow with IoT Core.

MikeDombo: It isn't valid to set the delta. This isn't allowed by iot core, and it isn't in our documentation either. We say that you can set desired and reported, that's all.

Nelsonochoam: Mmm even if we don't say it on the documentation doesn't mean someone might not be passing a delta for local updates. Out of extreme caution maybe then we could just omit sending it sending it to Iot Core or if we are going to do this maybe remove the delta when receiving the update and log a warning.

src/main/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandler.java
```diff
@@ -129,6 +131,14 @@ public UpdateThingShadowHandlerResponse handleRequest(UpdateThingShadowRequest r
                             .orElseThrow(() ->
                                     new InvalidRequestParametersException(ErrorMessage
                                             .createInvalidPayloadJsonMessage("")));
+
```
nikkhilmuthye, line 134: nit: Maybe we should add a TODO here to throw an exception when we want to move to a major version

src/test/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandlerTest.java
```diff
@@ -735,6 +738,22 @@ void GIVEN_bad_update_document_with_no_state_node_WHEN_handle_request_THEN_throw
         assertInvalidArgumentsErrorFromPayloadUpdate(null, badUpdateRequest, expectedErrorString, expectedErrorCode, context);
     }
 
+    @Test
+    void GIVEN_update_document_with_delta_WHEN_handle_request_THEN_delta_is_dropped_from_data() throws IOException, URISyntaxException {
+        byte[] badUpdateRequest =
```
nikkhilmuthye, line 743: We should capture the message sent to Synchandler and make sure that even that object does not have the `delta` node.

nikkhilmuthye (approved):

src/test/java/com/aws/greengrass/shadowmanager/ipc/UpdateThingShadowRequestHandlerTest.java
MikeDombo: This test covers it, without my change the test fails.

nikkhilmuthye: What this is testing is what the method is returning. I just want to completely make it future proof if we decide to use a different object we send to sync.

MikeDombo: Updated.

nikkhilmuthye (approved):

PR #179, opened 2023-04-12
fix: remove and recreate DB in the case of corruption

saranyailla:
**Issue #, if available:**

**Description of changes:**
When shadow manager DB is corrupted, the component goes into BROKEN state. The database file must be deleted for the component to create a new database and continue working. This change automates this deletion when the migration of DB fails due to corruption.  

**Checklist:**
 

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
```diff
@@ -57,21 +65,45 @@ public ShadowManagerDatabase(final Kernel kernel) {
...
+                }
+            } else {
+                throw flywaySqlException;
+            }
+        }
+    }
+
+    private void recreateDB(Flyway flyway) throws IOException {
+        logger.atDebug().kv("database-path", databasePath.toString()).log("Deleting the existing DB");
+        Files.deleteIfExists(databasePath);
```
MikeDombo, line 105: what about the other db files? There can be 3 or 4 files.

PR #180, opened 2023-04-12
fix: remove and recreate DB in the case of corruption

saranyailla:
**Issue #, if available:**

**Description of changes:**
When shadow manager DB is corrupted, the component goes into BROKEN state. The database file must be deleted for the component to create a new database and continue working. This change automates this deletion when the migration of DB fails due to corruption.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
@@ -8,6 +8,7 @@ package com.aws.greengrass.shadowmanager;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.shadowmanager.exception.ShadowManagerDataException;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.Getter;
 import lombok.Synchronized;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
```diff
@@ -57,21 +65,45 @@ public ShadowManagerDatabase(final Kernel kernel) {
...
+                }
+            } else {
+                throw flywaySqlException;
+            }
+        }
+    }
+
+    private void recreateDB(Flyway flyway) throws IOException {
+        logger.atDebug().kv("database-path", databasePath.toString()).log("Deleting the existing DB");
+        Files.deleteIfExists(databasePath);
```
MikeDombo, line 105: how about the other database files? The database can have 3 or 4 files.

saranyailla: We're setting up the database with `DEFRAG_ALWAYS=TRUE` which should ensure that only one db file exists. (Other files are tracelog and temp.db file). I'm deleting all the files in the workfolder that end with "db".

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
```diff
@@ -102,7 +101,15 @@ private void migrateDB(Flyway flyway) throws IOException, FlywayException {
 
     private void recreateDB(Flyway flyway) throws IOException {
         logger.atDebug().kv("database-path", databasePath.toString()).log("Deleting the existing DB");
-        Files.deleteIfExists(databasePath);
+        Files.list(databasePath).forEach((path -> {
```
MikeDombo, line 104: files.list gives you a stream that you MUST close. Use this with try-with-resources

src/main/java/com/aws/greengrass/shadowmanager/ShadowManagerDatabase.java
```diff
@@ -101,15 +102,18 @@ private void migrateDB(Flyway flyway) throws IOException, FlywayException {
...
-            if (path.endsWith("db")) {
+        try (Stream<Path> workPathFiles = Files.list(databasePath)) {
+            workPathFiles.forEach(path -> {
+                if (!path.endsWith("db")) {
+                    return;
+                }
                 try {
                     Files.deleteIfExists(path);
                 } catch (IOException e) {
                     throw new RuntimeException(e);
```
MikeDombo, line 113: this entire method can now throw an runtime exception, I think you need to catch it and handle it in this recreateDB method.

Nelsonochoam: Why just not throw a `ShadowManagerDataException` we are re wrapping it above in it. Also why not just wrap the entire thing in a try catch, catching IOException?. This method should just throw a `ShadowManagerDataException` instead of having to add try catch all over this method and the caller code

MikeDombo (approved):

Nelsonochoam (approved):

PR #192, opened 2023-09-19
fix: support empty state document

jcosentino11:
**Issue #, if available:**

**Description of changes:**
Support an empty shadow document,`{"state": {}}`.

According to https://docs.aws.amazon.com/iot/latest/developerguide/device-shadow-document.html#device-shadow-empty-fields, `{}` is a valid shadow document.

New unit and integration tests

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/model/ShadowState.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/model/ShadowState.java
@@ -58,6 +58,11 @@ public class ShadowState {
      */
     @SuppressWarnings({"PMD.ForLoopCanBeForeach", "PMD.NullAssignment"})
     public void update(JsonNode updatedStateNode) {
+        if (isNullOrMissing(updatedStateNode)) {
+            this.desired = null;
+            this.reported = null;
+            return;
+        }
         for (final Iterator<String> i = updatedStateNode.fieldNames(); i.hasNext(); ) {
             final String field = i.next();
             final JsonNode value = updatedStateNode.get(field);
--- a/src/main/resources/json/schema/update_payload_schema.json
+++ b/src/main/resources/json/schema/update_payload_schema.json
@@ -11,7 +11,7 @@
     },
     "state": {
       "description": "The state node containing the desired, reported, and delta shadow state.",
-      "type": "object",
+      "type": ["object", "null"],
       "properties": {
         "reported": {
           "description": "The shadow document state reported by the components.",
--- a/src/main/resources/json/schema/update_payload_schema.json
+++ b/src/main/resources/json/schema/update_payload_schema.json
@@ -32,6 +32,9 @@
         },
         {
           "required": [ "desired" ]
+        },
+        {
+          "additionalProperties": false
         }
       ]
     },
--- a/src/test/java/com/aws/greengrass/shadowmanager/util/JsonUtilTest.java
+++ b/src/test/java/com/aws/greengrass/shadowmanager/util/JsonUtilTest.java
@@ -59,6 +59,7 @@ class JsonUtilTest {
 
     @ParameterizedTest
     @ValueSource(strings = {
+            "{\"state\": null}",
             "{\"state\": {}}",
             "{\"state\": {\"desired\":" + NAME_A + ", \"reported\":" + NAME_B + ", \"delta\":" + NAME_A + "}}",
             "{\"state\": {\"desired\":" + NAME_A + ", \"reported\":" + NAME_A + "}}",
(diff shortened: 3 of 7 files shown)


MikeDombo (dismissed):

src/test/java/com/aws/greengrass/shadowmanager/util/JsonUtilTest.java
```diff
@@ -42,7 +42,6 @@ void setup() throws IOException {
     @ValueSource(strings={
             "{\"version\": 1}",
             "{}",
-            "{\"state\": {}}",
```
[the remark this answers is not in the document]
jcosentino11: According to docs I don't believe so, but I'll make sure to confirm with manual testing

saranyailla (approved):

src/test/java/com/aws/greengrass/shadowmanager/util/JsonUtilTest.java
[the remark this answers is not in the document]
jcosentino11: testing in IoT console and cli, updating a shadow with`null` is valid, but subsequent gets will return `{}`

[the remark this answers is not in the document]
MikeDombo: OK, so we need to support that too then.

And how about just `{}` without `{"state": <something>}`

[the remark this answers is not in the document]
jcosentino11: yep just added `null` support.  

{} without state is invalid:
```The Device Shadow couldn't be updated. Missing required node: state```

**Review by MikeDombo — APPROVED** [review]

File: src/main/resources/json/schema/update_payload_schema.json [hunk 2]
```diff
@@ -11,7 +11,7 @@
     },
     "state": {
       "description": "The state node containing the desired, reported, and delta shadow state.",
-      "type": "object",
+      "type": ["object", "null"],
```
MikeDombo, line 14: so we're considering these values as valid now, which is fine, but are we actually handling the values correctly?

What is our behavior after validating the data?

jcosentino11:
Adding more tests for local to cloud sync, realized that merging logic needs to be adjusted too

src/main/resources/json/schema/update_payload_schema.json
jcosentino11: yeah realized this after more testing, currently reviewing/updating merging logic to handle null/empty state

src/main/java/com/aws/greengrass/shadowmanager/model/ShadowState.java
```diff
@@ -58,6 +58,11 @@ public ShadowState deepCopy() {
      */
     @SuppressWarnings({"PMD.ForLoopCanBeForeach", "PMD.NullAssignment"})
     public void update(JsonNode updatedStateNode) {
+        if (isNullOrMissing(updatedStateNode)) {
+            this.desired = null;
+            this.reported = null;
```
MikeDombo, line 63: why are both being set to null?

jcosentino11: My understanding is that if shadow is being updated (either cloud or local) with "state: {}" or "state: null", then that signals we want to "clear" the document, so setting desired/reported to null

MikeDombo: Need to verify that. Updating the shadow state should only impact the reported state if the update is coming from the device, IMO.

jcosentino11: To clear just the reported or just the desired state, it'd be `{"state": {"reported": null}`, so sending a `{"state": {}}` seems pretty intentional for clearing the full shadow, although destructive.  I don't think we can truly disambiguate if devices intended to clear reported or desired, since customers can modify both reported or desired state (even if they shouldn't).

MikeDombo: OK, I see.

MikeDombo: And do we handle that case?

jcosentino11: Yeah null reported/desired is already expected/handled throughout the code

MikeDombo (approved):

saranyailla (approved):

PR #201, opened 2024-01-23
fix: retain null fields in sync node merger

sameerzuberi:
**Issue #, if available:**

**Description of changes:**

- Previously when a full sync was performed, `SyncNodeMerger` would remove any fields that were set to `null` from the shadow document. Then, once this document was merged during the local update, those removed fields were reintroduced since the merge was unaware they had already been set to `null` before. With this fix we now retain null fields inside of `SyncNodeMerger` so that they can be handled and deleted from the local shadow document during the following merges.
- Updated the integration test framework to allow for setting an initial local shadow state before starting the Nucleus. This allows us to simulate receiving cloud updates while the device is not yet running.

Fixes a bug where customers could delete fields in their cloud shadow while the device was offline, but after starting the device would not remove these fields from the local shadow. 

Added an integration test which fails without the change and passes with it.

`FullShadowSyncRequestTest.java` will need to be refactored in a future PR. With this change, any item removed from a shadow document is set to null in `SyncNodeMerger` and then handled accordingly in `UpdateThingShadowRequestHandler`. In the current unit tests, `UpdateThingShadowRequestHandler` has mocked behavior which causes the null values to not be handled as they would be in practice, so the null value has been added into the expected result for now, pending a refactor of the tests.

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/util/JsonMerger.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/util/JsonMerger.java
@@ -65,19 +65,6 @@ public final class JsonMerger {
                 + " Json objects whose underlying node are the same type."));
     }
 
-    private static void removeDeletedFields(final ObjectNode source, final ObjectNode patch) {
-        final Iterator<String> fieldNames = source.fieldNames();
-        while (fieldNames.hasNext()) {
-            final String field = fieldNames.next();
-            final JsonNode patchValue = patch.get(field);
-
-            // If the patch value is null then the field has been deleted, remove from source
-            if (isNullOrMissing(patchValue)) {
-                source.remove(field);
-            }
-        }
-    }
-
     private static void merge(final ObjectNode source, final ObjectNode patch) {
         final Iterator<String> fieldNames = patch.fieldNames();
         while (fieldNames.hasNext()) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/shadowmanager/util/JsonMerger.java
```diff
@@ -44,6 +44,9 @@ public static void merge(JsonNode source, final JsonNode patch) throws InvalidRe
             return;
         }
 
+        // If the source node contains keys not present in the patch node, those keys have been deleted
+        removeDeletedFields((ObjectNode) source, (ObjectNode) patch);
```
MikeDombo, line 48: I'm concerned that this method (merge) may be used by multiple code paths with different types of shadow documents. Whether the "patch" is truly a patch, or if it is supposed to be the complete shadow document.

Have you checked to see the code paths that this is used to ensure that we're doing the right thing?

If this method is used to merge in a delta change, then that should not result in fields being removed since the delta is only including changed fields rather than all fields.

sameerzuberi: That's something that just occurred to me also, going to dig into that a bit more to make sure this isn't breaking any other functionality

sameerzuberi: Ah looks like this path is used for other merges, I'll look into another approach for this

MikeDombo (approved):

jrcarbAWS (approved):

jcosentino11 (approved):

PR #204, opened 2024-05-21
fix: synchronize cloud data client operations properly

saranyailla:
**Issue #, if available:**

**Description of changes:**
Run mqtt callbacks in a separate thread to avoid a deadlock situation that happens when the Shadow manager component enters into RUNNING state before the MQTT client connection is successfully created acc to GG. 

Mqtt connect future will be completed with the client only after the first on connect callbacks are triggered. Shadow manager onConnect callback needs the client to be fully formed (connect future to be completed with the mqtt client) for it to use subscribe with it.  Hence, the subscriptions triggered from the callback timeout waiting for the client.

During SM start up, startSyncingShadows is called which calls [updateSubscriptions](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/315d1b9cdf7573f8ad4d995835320c5477f9e3ff/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L561) on the cloudDataClient. That spins up a new thread from the executor service pool which run this private synchronized  [updateSubscriptions](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/36bb2636d35b3dfd0719e12ab5d3f20562157056/src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java#L148) on the cloudDataClient. This runs [indefinitely](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/36bb2636d35b3dfd0719e12ab5d3f20562157056/src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java#L56) as mqtt subscribe op was never  [successful](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/36bb2636d35b3dfd0719e12ab5d3f20562157056/src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java#L182).
Now, mqtt callback thread is blocked at [updateSubscriptions](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/315d1b9cdf7573f8ad4d995835320c5477f9e3ff/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L561) in startSyncShadows because that method is also synchronized on the cloudDataClient instance and we can't have two synchronized methods interleaving  on the same instance.

More info:
When the MQTT client is created for the first time, onConnect (one-time) callbacks are run before the [connectFuture is completed with the client](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/a24441f08bf9e08553b8b029da171e702204bc32/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqtt5Client.java#L127-L132).  Only when these callbacks are completed, the `connectFuture` is completed. 

But, in the case where Shadow manager component enters into RUNNING state before the MQTT client connection is successfully created for the first time, [onConnectionResumed](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/bbb5cfa939a63ca2f2f0935f816ef1859cb12c7c/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L112-L117) callback is triggered when the mqtt client is created for the first time.  This callback uses subscribes to topics using mqtt client. However, in order to subscribe using the mqtt client, the [`connectFuture`](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/a24441f08bf9e08553b8b029da171e702204bc32/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqtt5Client.java#L282) should be fully completed resulting in a deadlock situation. 

The fix is to run the callback in a separate thread, so the `connectFuture` is completed without being blocked.  

**Checklist:**
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -111,17 +111,15 @@ public class ShadowManager extends PluginService {
     public final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
-            handleAsync(() -> stopSyncingShadows(true));
+            stopSyncingShadows(true);
         }
 
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
-            handleAsync(() -> {
-                if (inState(State.RUNNING)) {
-                    startSyncingShadows(
-                            StartSyncInfo.builder().startSyncStrategy(true).updateCloudSubscriptions(true).build());
-                }
-            });
+            if (inState(State.RUNNING)) {
+                handleAsync(() -> startSyncingShadows(
+                        StartSyncInfo.builder().startSyncStrategy(true).updateCloudSubscriptions(true).build()));
+            }
         }
 
         private void handleAsync(Runnable runnable) {
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -63,9 +63,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
@@ -104,10 +101,7 @@ public class ShadowManager extends PluginService {
     private final CloudDataClient cloudDataClient;
     private final MqttClient mqttClient;
     private final PubSubIntegrator pubSubIntegrator;
-    private final ExecutorService executorService;
     private final AtomicReference<Strategy> currentStrategy = new AtomicReference<>(DEFAULT_STRATEGY);
-    // This is used from within the mqtt thread only
-    private Future<?> mqttCallbackFuture = new CompletableFuture<>();
     public final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
@@ -116,20 +110,14 @@ public class ShadowManager extends PluginService {
 
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
+            // Make sure that it is non-blocking as it is run on mqtt event loop thread.
             if (inState(State.RUNNING)) {
-                handleAsync(() -> startSyncingShadows(
-                        StartSyncInfo.builder().startSyncStrategy(true).updateCloudSubscriptions(true).build()));
-            }
-        }
+                startSyncingShadows(StartSyncInfo.builder().startSyncStrategy(true)
+                        .updateCloudSubscriptions(true).build());
 
-        private void handleAsync(Runnable runnable) {
-            if (!mqttCallbackFuture.isDone() && !mqttCallbackFuture.isCancelled()) {
-                mqttCallbackFuture.cancel(true);
             }
-            mqttCallbackFuture = executorService.submit(runnable);
         }
     };
-
     private final CallbackEventManager.OnConnectCallback onConnect = callbacks::onConnectionResumed;
 
     @Getter(AccessLevel.PUBLIC)
@@ -160,18 +148,23 @@ public class ShadowManager extends PluginService {
      * @param cloudDataClient             the data client subscribing to cloud shadow topics
      * @param mqttClient                  the mqtt client connected to IoT Core
      * @param direction                   The sync direction
-     * @param executorService             Executor service
      */
     @SuppressWarnings("PMD.ExcessiveParameterList")
     @Inject
     public ShadowManager(
             Topics topics,
-            ShadowManagerDatabase database, ShadowManagerDAOImpl dao,
-            AuthorizationHandlerWrapper authorizationHandlerWrapper, PubSubClientWrapper pubSubClientWrapper,
-            InboundRateLimiter inboundRateLimiter, DeviceConfiguration deviceConfiguration,
-            ShadowWriteSynchronizeHelper synchronizeHelper, IotDataPlaneClientWrapper iotDataPlaneClientWrapper,
-            SyncHandler syncHandler, CloudDataClient cloudDataClient, MqttClient mqttClient, DirectionWrapper direction,
-            ExecutorService executorService) {
+            ShadowManagerDatabase database,
+            ShadowManagerDAOImpl dao,
+            AuthorizationHandlerWrapper authorizationHandlerWrapper,
+            PubSubClientWrapper pubSubClientWrapper,
+            InboundRateLimiter inboundRateLimiter,
+            DeviceConfiguration deviceConfiguration,
+            ShadowWriteSynchronizeHelper synchronizeHelper,
+            IotDataPlaneClientWrapper iotDataPlaneClientWrapper,
+            SyncHandler syncHandler,
+            CloudDataClient cloudDataClient,
+            MqttClient mqttClient,
+            DirectionWrapper direction) {
         super(topics);
         this.database = database;
         this.authorizationHandlerWrapper = authorizationHandlerWrapper;
@@ -192,7 +185,6 @@ public class ShadowManager extends PluginService {
         this.pubSubIntegrator = new PubSubIntegrator(pubSubClientWrapper, deleteThingShadowRequestHandler,
                 updateThingShadowRequestHandler, getThingShadowRequestHandler);
         this.direction = direction;
-        this.executorService = executorService;
     }
 
     private void registerHandlers() {
--- a/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
@@ -63,6 +63,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -101,22 +104,34 @@ public class ShadowManager extends PluginService {
...
+                if (inState(State.RUNNING)) {
+                    startSyncingShadows(
+                            StartSyncInfo.builder().startSyncStrategy(true).updateCloudSubscriptions(true).build());
+                }
+            });
+        }
 
+        private void handleAsync(Runnable runnable) {
+            if (!mqttCallbackFuture.isDone() && !mqttCallbackFuture.isCancelled()) {
+                mqttCallbackFuture.cancel(true);
```
junfuchen99, line 129: is it a problem if mqtt disconnects and connects quickly - `stopSyncingShadows` is in-progress and cancelled prematurely before the next `startSyncingShadows`

junfuchen99: feels like maybe stopSyncingShadows shouldn't be async

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
```diff
@@ -101,22 +104,34 @@ public class ShadowManager extends PluginService {
...
+            handleAsync(() -> stopSyncingShadows(true));
         }
 
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
-            if (inState(State.RUNNING)) {
-                startSyncingShadows(StartSyncInfo.builder().startSyncStrategy(true)
-                        .updateCloudSubscriptions(true).build());
+            handleAsync(() -> {
+                if (inState(State.RUNNING)) {
```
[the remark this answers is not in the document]
jcosentino11: we'd also need to check inside the handler too async handler will run at some future point

jcosentino11:
Can we accomplish this without adding more threads?  In startSyncingShadows, we can check if connected in a non-blocking way `mqttClient.getMqttOnline().get()`.  and for stopSyncingShadows,  maybe we can find a way to not `waitForSyncEnd();` in this case

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
saranyailla: https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/36bb2636d35b3dfd0719e12ab5d3f20562157056/src/main/java/com/aws/greengrass/shadowmanager/sync/strategy/BaseSyncStrategy.java#L195  there may be some waiting while stopping as it is synchronized. 

But, I'm updating this as it feels like we need to wait and there's not anything we're doing with the mqttclient itself.

src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java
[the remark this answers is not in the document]
saranyailla: Updated it as suggested as we can avoid starting a new thread if the component in not in RUNNING state. 
And to Joe's point, we do check if the component is in `RUNNING` in `startSyncingShadows` as well. So, it should be covered when the handler runs in future.

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -145,7 +145,7 @@ public synchronized void updateSubscriptions(Set<Pair<String, String>> shadowSet
      * @param updateTopics Set of update shadow topics to subscribe to
      * @param deleteTopics Set of delete shadow topics to subscribe to
      */
-    private synchronized void updateSubscriptions(Set<String> updateTopics, Set<String> deleteTopics) {
+    private void updateSubscriptions(Set<String> updateTopics, Set<String> deleteTopics) {
```
jcosentino11, line 148: needs a little more refactoring, `subscribedUpdateShadowTopics` and `subscribedDeleteShadowTopics` need synchronization now, maybe they can be passed in as params to this method or something

jcosentino11: and with this i don't believe we'd need `handleAsync` in the onConnectionResumed anymore

saranyailla: But, we are doing some work on mqtt event loop thread which is prone to deadlocks/blocking conditions due to some future changes in any part of the shadow manager code.

saranyailla: Both `unsubscribeToShadows` and `subscribeToShadows` are private methods that are `synchronized`. This is still good as they are not blocking like `updateSubscriptions` by retrying indefinitely. Also, `updateSubscriptions` is called from a method that is `synchronized`.

jcosentino11: > But, we are doing some work on mqtt event loop thread which is prone to deadlocks/blocking conditions due to some future changes in any part of the shadow manager code.   

Which is exactly why we should err towards reducing async/locks where possible, and why with future changes we need to remain vigilant. I'm worried that making this async will cover up more issues.

MikeDombo: The MQTT client now offers async APIs, since Nucleus 2.10 :)

jcosentino11: > Both `unsubscribeToShadows` and `subscribeToShadows` are private methods that are `synchronized`. This is still good as they are not blocking like `updateSubscriptions` by retrying indefinitely. Also, `updateSubscriptions` is called from a method that is `synchronized`. 

Right, the issue is that internal updateSubscriptions is sharing the lock with the other methods that don't block

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -145,74 +120,78 @@ public synchronized void updateSubscriptions(Set<Pair<String, String>> shadowSet
...
-                // if interrupted then handle
-                if (Thread.currentThread().isInterrupted()) {
-                    Thread.currentThread().interrupt();
-                    logger.atError()
-                            .setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code())
-                            .log("Failed to update subscriptions");
-                    return false;
-                }
+                    // if interrupted then handle
+                    if (Thread.currentThread().isInterrupted()) {
```
jcosentino11, line 164: i realize this is pre-existing, but let's just get rid of this whole block.  the subscribe/unsubscribe methods above will throw InterruptedException.  And returning `false` here on interrupt is wrong imo, the operation *did* succeed at this point.   Let's also adjust the logs while we're at it:  interruption isn't an error, it's part of the expected behavior/flow, either shutdown or we're doing a new task, so "failed to update" isn't true

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -48,9 +49,9 @@ public class CloudDataClient {
     private final MqttClient mqttClient;
     private final ExecutorService executorService;
     @Getter(AccessLevel.PACKAGE)
-    private final Set<String> subscribedUpdateShadowTopics = new HashSet<>();
+    private final Set<String> subscribedUpdateShadowTopics = ConcurrentHashMap.newKeySet();
```
jcosentino11, line 52: don't need this anymore since `subscribedUpdateShadowTopics` and `subscribedDeleteShadowTopics` are now fully protected by `subscriptionLock` now

jcosentino11 (commented):
lgtm, just some small cleanup suggestions

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
saranyailla: Yes, with all the current changes we got, but shouldn't hurt to keep it that way.

src/main/java/com/aws/greengrass/shadowmanager/sync/CloudDataClient.java
```diff
@@ -145,74 +120,71 @@ public synchronized void updateSubscriptions(Set<Pair<String, String>> shadowSet
...
-        } catch (Exception e) { // NOPMD - thrown by RetryUtils.runWithRetry()
-            logger.atError()
-                    .setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code())
-                    .setCause(e)
-                    .log("Failed to update subscriptions");
-            return;
-        }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                logger.atError().setEventType(LogEvents.CLOUD_DATA_CLIENT_SUBSCRIPTION_ERROR.code()).setCause(e)
```
jcosentino11, line 172: nit: info, or even debug, and don't really need cause anymore

src/test/java/com/aws/greengrass/shadowmanager/sync/CloudDataClientTest.java
```diff
@@ -290,6 +290,7 @@ void GIVEN_100_synced_shadows_WHEN_unsubscribeForAllShadowsTopics_THEN_unsubscri
         }
 
         cloudDataClient.unsubscribeForAllShadowsTopics();
+        TimeUnit.MILLISECONDS.sleep(5000);
```
jcosentino11, line 293: what's this for?

jcosentino11 (approved):

src/test/java/com/aws/greengrass/shadowmanager/sync/CloudDataClientTest.java
saranyailla: `cloudDataClient.unsubscribeForAllShadowsTopics();` now runs in a separate thread with the current changes.

jcosentino11: yeah ideally we have a way of waiting via a latch or something

junfuchen99 (approved):

Issue #212, opened 2025-09-11
(ShadowManager): NullPointerException on shadow desired state field deletion


edreanernst:
**Describe the bug**
When the local shadow on a Greengrass deployment is updated with a desired state of null for a nested field (the accepted method to remove the field's state from the shadow), Shadow Manager throws a NullPointerException, presumably when it attempts to calculate a delta.

**To Reproduce**
First, send a payload to Shadow Manager such as:
`{"state": {"desired": {"mystate": {"mysub": "123"}}}}`
Then send a payload after that:
`{"state": {"desired": {"mystate": {"mysub": null}}}}`

**Expected behavior**
The expectation is that the field `mysub` in my example should be removed from the shadow document without any errors.

**Actual behavior**
A NullPointerException is thrown and continues to be thrown whenever other subsequent shadow updates are attempted until the `null` is cleared from the document by another update to the shadow for field `mysub` with a value other than `null`.

The NullPointerException is:

```
[ERROR] (AwsEventLoop 1) com.aws.greengrass.ipc.common.ExceptionUtil: Unhandled exception in IPC. {}
java.lang.NullPointerException
        at com.aws.greengrass.shadowmanager.model.ShadowStateMetadata.buildMetadata(ShadowStateMetadata.java:294)
        at com.aws.greengrass.shadowmanager.model.ShadowStateMetadata.getDeltaMetadata(ShadowStateMetadata.java:254)
        at com.aws.greengrass.shadowmanager.model.ShadowDocument.getDelta(ShadowDocument.java:202)
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler.publishDeltaMessage(UpdateThingShadowRequestHandler.java:309)
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler.lambda$handleRequest$2(UpdateThingShadowRequestHandler.java:236)
        at com.aws.greengrass.ipc.common.ExceptionUtil.translateExceptions(ExceptionUtil.java:33)
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowRequestHandler.handleRequest(UpdateThingShadowRequestHandler.java:94)
        at com.aws.greengrass.shadowmanager.ipc.UpdateThingShadowIPCHandler.handleRequest(UpdateThingShadowIPCHandler.java:83)
[...]

PR #213, opened 2025-10-03
[Fix] Leave empty objects in metadata

kstribrnAmzn:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-shadow-manager/issues/212

**Description of changes:**
This change updates how Shadow document metadata is calculated. For a simple document updates which remove only top key value pairs, the behavior will not change. For a more complicated document where nested objects are removed, the behavior is updated to keep the empty parent object (which matches IoT Shadow). 

By removing empty objects from the metadata the shadow manager is calculating the ShadowDocMetadata.desired (or reported) field to null which results in a later NullPointerException when building the updated metadata later.

Added unit tests as well as performed manual testing against the case mentioned in the issue.

It's useful to know how Jackson JsonNode's function

**Checklist:**
 - [N/A] Updated the README if applicable
 - [Pending] Updated or added new integration tests
 - [N/A] If your code makes a remote network call, it was tested with a proxy
 
### DIFF
--- a/src/main/java/com/aws/greengrass/shadowmanager/model/ShadowStateMetadata.java
+++ b/src/main/java/com/aws/greengrass/shadowmanager/model/ShadowStateMetadata.java
@@ -70,8 +70,7 @@ public class ShadowStateMetadata {
      */
     @SuppressWarnings("PMD.NullAssignment")
     public JsonNode update(JsonNode patch, ShadowState state) {
-        // Create the patch metadata tree. This will transform nulls to metadata nodes.
-        final JsonNode metadataPatchWithRemovedFields = createMetadataPatch(patch, false);
+        // The persisted metadata should have all removed fields (aka value null) actually removed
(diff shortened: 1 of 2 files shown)


kstribrnAmzn:
Some manual testing logs snippets. In this case I created a shadow document and then set a field to null to remove it. For the case of removing a simple value node (aka JSON value) the shadow document is updated to 
```
GetThingShadowResponse(payload=b'{"state":{},"metadata":{},"version":24,"timestamp":1759530985}') : {'payload': '<payload>'}.
```

Now for the case of removing a child key-value pair, it is shown that the parent structure maintains.
```
GetThingShadowResponse(payload=b'{"state":{"desired":{"created":{}},"delta":{"created":{}}},"metadata":{"desired":{"created":{}}},"version":22,"time
stamp":1759530294}') : {'payload': '<payload>'}.

```

saranyailla (dismissed):

saranyailla (approved):

